PR #3, opened 2017-01-17
implemented new methods in StreamParser

adamiwaniuk:


MaciekPytel:
This code seem both non-trivial and easily unittestable, so it would be good to add unittests for it (in a separate PR later)

include/parser/stream.h
```diff
@@ -125,6 +234,19 @@ class StreamParser {
...
 
+  std::vector<uint8_t> getBytesUntil(const QString &name, uint8_t termination,
+                                     bool include_termination = true) {
+    auto data =
+        getDataUntil(name, data::RepackFormat{data::RepackEndian::LITTLE, 8},
+                     data::BinData::fromRawData(8, {termination}),
+                     data::FieldHighType(), include_termination);
+    std::vector<uint8_t> res;
+    for (size_t i = 0; i < data.size(); i++) {
+      res.push_back(data.element64(i));
```
MaciekPytel: Is this really how we're supposed to use BinData? Feels like we're doing a lot of unnecessary stuff just to pack BinData.rawData into uint64_t, so we can take it out again. 

Also I would expect to get compiler warnings in here (implicit cast from uint64_t to uint8_t); at the very least can we add an explicit cast?

include/parser/stream.h
```diff
@@ -134,26 +256,40 @@ class StreamParser {
...
+  std::vector<uint16_t> get16(const QString &name, uint64_t num,
+                              data::RepackEndian endian) {
     std::vector<uint16_t> res;
-    auto data = getData(
-      name, data::RepackFormat{data::RepackEndian::LITTLE, 16}, num,
-      data::FieldHighType());
+    auto data = getData(name, data::RepackFormat{endian, 16}, num,
+                        data::FieldHighType());
     for (size_t i = 0; i < data.size(); i++) {
       res.push_back(data.element64(i));
```
MaciekPytel: Same as above - can we at least add an explicit cast in here?

include/parser/stream.h
adamiwaniuk: the same code is right now in getBytes, so maybe latter we can separate common code into another function and make some improvments here

include/parser/stream.h
adamiwaniuk: actually this code is not something added in this PR,  changed

PR #8, opened 2017-01-17
Wip/altran/network communication

altran01:


include/db/object.h
```diff
@@ -60,6 +63,8 @@ class LocalObject : public QEnableSharedFromThis<LocalObject> {
   virtual dbif::ObjectType type() const = 0;
   QString name() const { return name_; }
   QString comment() const { return comment_; }
+  uint64_t id() const { return id_; }
+  const QSet<PLocalObject> &children() { return children_; }
```
MaciekPytel: nitpick: in other places you do const X& x(), not const X &x() (I like the first one more)

include/network/server.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <QtNetwork/QTcpServer>
+
+#include "db/types.h"
+#include "network.pb.h"
+
+namespace veles {
+namespace db {
+
+class NetworkServer : public QObject {
```
MaciekPytel: Do we want to use Qt on server side? I'm not saying we don't but we should discuss.

Regardless I think it's ok for POC.

MaciekPytel:
I think we need at least an option to create and read blobs (perhaps read file blobs by passing just filepath, but that should be somehow encapsulated).

python/README.rst
```diff
@@ -0,0 +1,47 @@
...
+This library enables you to operate on veles objects tree from python.
+It connects to running veles instance via network.
+
+Installing
+----------
+
+First you need to copy ``network_pb2.py`` to ``veles`` directory (this file is
+generated during Veles compilation).
+
+Then you need to run ``python setup.py install`` (or ``develop`` based on your preferences)
```
MaciekPytel: Shouldn't setup.py do the copy network_pb2.py part?

python/veles/objects.py
```diff
@@ -0,0 +1,100 @@
...
+        self._proto_obj.chunk_type = chunk_type
+
+    @property
+    def items(self):
+        return self._proto_obj.items
+
+    # TODO setting items
+
+    def __repr__(self):
+        return 'id: %s name: %s' % (self.id, self.name)
```
MaciekPytel: use .format instead of % formatting

python/veles/veles_api.py
```diff
@@ -0,0 +1,125 @@
...
+        id_path = id_path[:] + [res.id]
+        obj = objects.LocalObject(res, id_path)
+        for child in res.children:
+            child_obj = self._prepare_object(child, id_path, obj)
+            obj.children.append(child_obj)
+        return obj
+
+    def list_files(self):
+        req = network_pb2.Request()
+        req.type = 1
```
MaciekPytel: That really should come from an enum or map of some sort

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+  connect(client_connection, &QAbstractSocket::disconnected,
+          client_connection, &QObject::deleteLater);
+  connect(client_connection, &QIODevice::readyRead, [this, client_connection] () {
+    readMessage(client_connection);
+  });
+}
+
+void NetworkServer::readMessage(QTcpSocket *client_connection) {
+  int32_t msg_len;
+  if (client_connection->bytesAvailable() < static_cast<int32_t>(sizeof(msg_len))) {
```
MaciekPytel: Why is message length a signed integer? Someone can easily craft a packet with this being a negative number.

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+    return;
+  }
+
+  client_connection->peek(reinterpret_cast<char*>(&msg_len), sizeof(msg_len));
+  msg_len = qFromLittleEndian(msg_len);
+  if (client_connection->bytesAvailable() >=
+      static_cast<int32_t>(sizeof(msg_len)) + msg_len) {
+    QScopedArrayPointer<char> message(new char[msg_len]);
```
MaciekPytel: Also maybe we want a maximum on msg_len?

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+    for (auto child : target_object->children()) {
+      if (child->id() == req.id(j)) {
+        if (child->type() == dbif::FILE_BLOB) {
+          blob = child;
+        }
+        found = true;
+        target_object = child;
+        break;
+      }
+    }
```
MaciekPytel: Maybe root (or someone) should have a set of all nodes, to avoid bruteforcing our way through the tree?

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+
+    if (!found) {
+      resp.set_ok(false);
+      resp.set_error_msg("Bad ID provided.");
+      return;
+    }
+  }
+
+  switch (req.type()) {
+  case 1:
```
MaciekPytel: enum

include/db/object.h
```diff
@@ -33,6 +35,8 @@ class LocalObject : public QEnableSharedFromThis<LocalObject> {
   Universe *db_;
   QString name_;
   QString comment_;
+  static std::atomic<uint64_t> static_id_;
+  uint64_t id_;
```
MaciekPytel: What do you think about also adding version field? We don't need it before we start modifying objects over the network, but once we do allow that we need a way to defend against race conditions on updates.

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+  default:
+    resp.set_ok(false);
+    resp.set_error_msg("Unknown request type.");
+    break;
+  }
+}
+
+void NetworkServer::sendResponse(QTcpSocket *client_connection,
+                                 network::Response &resp) {
+  int32_t resp_len = resp.ByteSize(), resp_len_send = qToLittleEndian(resp_len);
```
MaciekPytel: Once again, why is this a signed int?

python/README.rst
altran01: How would you like to specify directory where veles (and network_pb2.py) is build - as parameter to setup.py script ?

python/veles/veles_api.py
altran01: Will do (don't really know why I didn't use enum here...)

src/network/server.cc
altran01: ByteSize returned signed int so automatically stuck with it - however you are right we should use unsigned here - will change.

include/network/server.h
altran01: Currently we use Qt on "server" side (and thanks to it I didn't have to worry about portability of network code) so I decided to go with it for now.

altran01:
I was thinking about reading blobs option, but decided to add in another patch (which isn't yet created), however I can add it here if you want.

include/db/object.h
altran01: Yes it will be required if we allow edition - however that is not exactly the case currently (application ui only allows to edit chunk comment - which I didn't even implement in this network communication) - I think it would be better to add it when we really gonna use it.

src/network/server.cc
altran01: I was thinking about it - I even wrote a comment about it in .proto file (random access to objects on server side) but didn't implement it in this PoC version - I can add it if you want.

src/network/server.cc
altran01: Ah seems I've forgotten about it.

python/README.rst
MaciekPytel: Good point. Not sure what the right answer is, but maybe we can have cmake create a softlink to build directory or write the path somewhere specific?

include/db/object.h
MaciekPytel: fair enough

MaciekPytel:
I think it's pointless to have a functionality that can't be actually used for anything. And I can't imagine a use-case that doesn't require access to the data itself. So I think we need a way to read / create / delete blobs, to actually do anything useful at all. 

Personally I think we need at least blob-related stuff and a working example of python script (for parsing something simple like a bitmap, or just cutting ELF into sections) to make this a complete POC. What do you think @adamiwaniuk @koriakin @jacek-izykowski?

python/README.rst
altran01: Links would be a problem on Windows (by default only Administrator can create them and I don't requiring "sudo" to compile Veles would be good idea) - I'll think about alternatives.

MaciekPytel: Once again a good point. Maybe we can just write a text file with path or something?

src/network/server.cc
```diff
@@ -0,0 +1,238 @@
...
+  QScopedArrayPointer<char> response_content(new char[resp_len]);
+  resp.SerializeToArray(&response_content[0], resp_len);
+  int written = 0, total_written = 0;
+  while (total_written < static_cast<int32_t>(sizeof(resp_len_send))) {
+    written = client_connection->write(
+          ((const char*)&resp_len_send)+total_written,
```
MaciekPytel: formatting

altran01:
@MaciekPytel I think I taken all your comments into account.

include/network/server.h
```diff
@@ -0,0 +1,63 @@
...
+
+private slots:
+  void handle();
+  void readMessage(QTcpSocket *client_connection);
+
+private:
+  PLocalObject root_;
+  QTcpServer *tcp_server_;
+
+  static const uint32_t MAX_MSG_LEN = 1024*1024*16;
```
MaciekPytel: We use different const name convention elsewhere (k_max_msg_len_)

src/network/server.cc
```diff
@@ -0,0 +1,267 @@
...
+        found = true;
+        target_object = child;
+        break;
+      }
+    }
+
+    if (!found) {
+      resp.set_ok(false);
+      resp.set_error_msg("Bad ID provided.");
+      break;
```
MaciekPytel: Should we still go through the following switch if an error happened here?

python/examples/parse_bmp.py
```diff
@@ -0,0 +1,176 @@
...
+import veles.veles_api as vapi
+from veles import objects
+
+
+class BmpParser(object):
+    # TODO change a lot of chunks to items when they are supported
+    # and add some comments to fileds
+    # parse more variations of bitmap
+    def __init__(self):
+        self.client = vapi.VelesApi('127.0.0.1', 3135)
```
MaciekPytel: ip & port could be init params with defaults

src/network/server.cc
altran01: Ups, will fix.

python/examples/parse_bmp.py
```diff
@@ -0,0 +1,176 @@
...
+                self.parse_bmp(file)
+
+    def create_chunk(self, parent, name,
+                     start=None, end=None, size=None, comment=None):
+        new_obj = objects.LocalObject()
+        new_obj.name = name
+        if start is None:
+            new_obj.chunk_start = self.current_start
+        else:
+            new_obj.chunk_start = start
```
MaciekPytel: nitpicking: I think default arguments look prettier if you do

```
if start is None:
     start = self.current_start
new_obj.chunk_start = start
```

python/examples/parse_bmp.py
```diff
@@ -0,0 +1,176 @@
...
+        self.current_start = 0
+
+    def parse_files(self):
+        files = self.client.list_files()
+        for file in files:
+            if file.name.endswith('.bmp'):
+                self.parse_bmp(file)
+
+    def create_chunk(self, parent, name,
+                     start=None, end=None, size=None, comment=None):
```
MaciekPytel: Maybe it would be nice to later add this method to "standard" veles library? Seems quite convenient.

src/network/server.cc
```diff
@@ -109,18 +116,34 @@ void NetworkServer::createChunk(PLocalObject target_object, PLocalObject blob,
...
     return;
   }
   target_object->kill();
   resp.set_ok(true);
 }
 
-void NetworkServer::handleRequest(network::Request &req, network::Response &resp) {
+void NetworkServer::getBlobData(PLocalObject target_object, QTcpSocket *client_connection, network::Response &resp) {
+  if (target_object->type() != dbif::FILE_BLOB ||
+      target_object->type() == dbif::SUB_BLOB) {
```
MaciekPytel: This if is wierd - in particular second condition implies the first one, no?

src/network/server.cc
```diff
@@ -109,18 +116,34 @@ void NetworkServer::createChunk(PLocalObject target_object, PLocalObject blob,
...
+    resp.set_ok(false);
+    resp.set_error_msg("Unsupported object type to get file data.");
+    sendResponse(client_connection, resp);
+    return;
+  }
+  resp.set_ok(true);
+  sendResponse(client_connection, resp);
+
+  auto blob = target_object.dynamicCast<DataBlobObject>();
```
MaciekPytel: What's the point of dynamic cast if we don't check for nullptr?

python/README.rst
```diff
@@ -0,0 +1,51 @@
...
+  new_obj = objects.LocalObject()
+  new_obj.name = 'custom chunk'
+  new_obj.chunk_start = 0x10
+  new_obj.chunk_end = 0x20
+  c.create_chunk(files[0], new_obj)
+
+Delete chunk:
+::
+
+  c.delete_object(files[0].children[0])
```
MaciekPytel: Can we add explicit information to check out examples (and their location)?

python/veles/objects.py
```diff
@@ -0,0 +1,102 @@
...
+    # for now it's just a single class for all types of object
+    # when we have more final database schema we need to do it in better way
+    def __init__(self, proto_obj=None, id_path=None, parent=None):
+        if proto_obj is None:
+            self._proto_obj = network_pb2.LocalObject()
+        else:
+            self._proto_obj = proto_obj
+        self.parent = parent
+        if id_path is None:
+            self.id_path = []
```
MaciekPytel: Shouldn't that be based on parent's id_path?

python/veles/veles_api.py
```diff
@@ -108,7 +109,7 @@ def get_chunk_tree(self, obj):
             obj.children.append(res)
         return results
 
-    def create_chunk(self, obj, new_obj):
+    def raw_create_chunk(self, obj, new_obj):
```
MaciekPytel: maybe this should start with underscore?

PR #11, opened 2017-01-19
Tabs in main window can always be reordered.

jacek-izykowski:
MainWindowWithDetachableDockWidgets::dock_widgets_with_no_title_bars_ doesn't affect possibility to reorder tabs.

### DIFF
--- a/src/ui/veles_mainwindow.cc
+++ b/src/ui/veles_mainwindow.cc
@@ -299,6 +299,12 @@ TabBarEventFilter::TabBarEventFilter(QObject* parent) :
     drag_init_pos_(0, 0) {
 }
 
+void TabBarEventFilter::tabMoved(int from, int to) {
+  if (dragged_tab_bar_) {
+    dragged_tab_index_ = dragged_tab_bar_->currentIndex();
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/ui/veles_mainwindow.cc
```diff
@@ -335,8 +344,15 @@ bool TabBarEventFilter::eventFilter(QObject *watched, QEvent *event) {
...
-        > QApplication::startDragDistance()) {
+    bool horizontal_tabs =
+        dragged_tab_bar_->shape() == QTabBar::RoundedNorth
+        || dragged_tab_bar_->shape() == QTabBar::RoundedSouth
+        || dragged_tab_bar_->shape() == QTabBar::TriangularNorth
+        || dragged_tab_bar_->shape() == QTabBar::TriangularSouth;
+
+    if ((horizontal_tabs ? (event->pos() - drag_init_pos_).y()
+        : (event->pos() - drag_init_pos_).x())
+        > 5 * QApplication::startDragDistance()) {
```
MaciekPytel: I think this value (5 * startDragDistance) may require tuning later, so would be nice to have it on a const, rather than hardcode it here.

PR #12, opened 2017-01-19
Remove not (yet) implemented actions from toolbars/menus.

jacek-izykowski:


altran01:
I didn't get chance to review this code before it was merged - there is still at least one place with not yet implemented feature - find/replace - whole replace functionality isn't currently implemented.

PR #16, opened 2017-01-20
added parsers generated from kaitai ksy files

adamiwaniuk:


altran01:
Approve since there isn't a lot of not generated code (which I didn't read itself).

PR #18, opened 2017-01-20
Dock widgets splitting (available through context menu).

jacek-izykowski:


src/ui/veles_mainwindow.cc
```diff
@@ -647,6 +702,21 @@ QTabBar* MainWindowWithDetachableDockWidgets::dockWidgetToTab(
   return nullptr;
 }
 
+void MainWindowWithDetachableDockWidgets::splitDockWidget2(
+    QDockWidget* first, QDockWidget* second, Qt::Orientation orientation) {
+#ifdef SPLIT_DOCK_WIDGET_ACTUALLY_WORKS_AS_ONE_COULD_EXPECT
```
altran01: This isn't defined anywhere and I don' think this is some built in define on any of our platforms.

jacek-izykowski: Of course it's not a built in define because QMainWindow::splitDockWidget doesn't work satisfactory on any platform.

altran01: Then unless you plan on defining it, I don't think it should be an ifdef - it creates confusion that sometimes the behaviour differs, while it is just a rant that could be contained to comment.

jacek-izykowski: I do agree. Fixed.

src/ui/veles_mainwindow.cc
```diff
@@ -839,6 +908,31 @@ bool MainWindowWithDetachableDockWidgets::event(QEvent* event) {
...
+    removeDockWidget(dock_widget);
+  }
+  tabified_docks.removeOne(second);
+
+  second->show();
+  splitDockWidget(first, second, orientation);
+
+  for (auto dock_widget : tabified_docks) {
+    dock_widget->show();
+    if (dock_widget != second) {
```
MaciekPytel: How can dock_widget == second happen? You explicitly remove second a few lines up. Or can the splitDockWidget() append it back?

PR #19, opened 2017-01-20
moved sort to base class [jira EAX-174]

adamiwaniuk:


src/ui/fileblobitem.cc
```diff
@@ -47,8 +53,28 @@ void FileBlobItem::removingChildrenHandle(FileBlobItem *item, bool before) {
   emit removingChildren(item, before);
 }
 
+bool compareItems(FileBlobItem *a, FileBlobItem *b) { return *a < *b; }
+
+bool FileBlobItem::sortChildren() {
+  if (std::is_sorted(children_.begin(), children_.end(), compareItems)) {
+    return false;
+  }
+  qSort(children_.begin(), children_.end(), compareItems);
```
MaciekPytel: Why do we use std::is_sorted() and qSort in the same function??

adamiwaniuk: fixed

PR #20, opened 2017-01-20
don't scrroll to selected chunk if at least one byte is already visib…

adamiwaniuk:
…le [jira EAX-173]


altran01:
This fixes the problem I was having with selecting part of chunk, but I'm wondering whether should we preserve the old behaviour in some fashion - i.e single click works like after this change, but double click scrolls to beginning of chunk even if we see some other part of it.

adamiwaniuk:
double click right now goes inside chunk, and i expect in this case that we will go inside in exactly the same position (we want to see what is exacty on lower layer in clicekd byte). We can think about some keyboard shortcut to do this, but this is separate feature.

PR #21, opened 2017-01-20
Added new icons for different visualisation modes

MaciekPytel:
Also moved getColoredIcon function to utils and
reworked it a bit. Still it's kinda crap, end
result should be more antialiased.


src/visualisation/panel.cc
```diff
@@ -249,20 +250,26 @@ QBoxLayout* VisualisationPanel::prepareVisualisationOptions() {
 void VisualisationPanel::initOptionsPanel() {
   options_layout_ = new QVBoxLayout;
 
+  QColor icon_color = palette().color(QPalette::WindowText);
   digram_action_ =
-      new QAction(QIcon(":/images/nginx2d_32.png"), tr("&Digram"), this);
```
altran01: I think this icon isn't used anywhere any more and can be removed?

MaciekPytel: good point, will do

MaciekPytel:
@jacek-izykowski it's not dependant on the method really, it was just drawn to look somewhat ok, with the old method. A proper fix would be to just draw a new, nicer cylinder instead of trying to salvage old image. I can try and do that.

PR #33, opened 2017-01-30
Added separate python classes for blobs and chunks

MaciekPytel:
A whole bunch of related refactors in python API.
Updated examples/parse_bmp.py accordingly.

### DIFF
--- a/python/examples/parse_bmp.py
+++ b/python/examples/parse_bmp.py
@@ -20,149 +20,119 @@ class BmpParser(object):
                 self.parse_bmp(file)
 
     def parse_bmp(self, file):
-        self.client.get_blob_data(file)
-        if file.data[:2] != b'BM':
+        if file[:2] != b'BM':
             # unsupported file type
             return
-        file_size = struct.unpack('<I', file.data[2:6])[0]
-        dib_size = struct.unpack('<I', file.data[14:18])[0]
-        pix_start = struct.unpack('<I', file.data[10:14])[0]
+        file_size = struct.unpack('<I', file[2:6])[0]
+        dib_size = struct.unpack('<I', file[14:18])[0]
+        pix_start = struct.unpack('<I', file[10:14])[0]
 
-        bitmap = self.client.create_chunk(file, 'bitmap',
-                                          size=file_size, update_cursor=False)
+        bitmap = file.create_chunk('bitmap', size=file_size)
 
-        bitmap_header = self.client.create_chunk(
-            bitmap, 'Bitmap file header', end=14, update_cursor=False)
-        self.client.create_chunk(bitmap_header, 'FileType', size=2)
-        self.client.create_chunk(bitmap_header, 'FileSize', size=4)
-        self.client.create_chunk(
-            bitmap_header, 'Reserved1', size=2,
+        bitmap_header = bitmap.create_chunk('Bitmap file header', size=14)
+        bitmap_header.create_chunk('FileType', size=2)
+        bitmap_header.create_chunk('FileSize', size=4)
+        bitmap_header.create_chunk(
+            'Reserved1', size=2,
             comment='value depends on the application that creates the image')
-        self.client.create_chunk(
-            bitmap_header, 'Reserved2', size=2,
+        bitmap_header.create_chunk(
+            'Reserved2', size=2,
             comment='value depends on the application that creates the image')
-        self.client.create_chunk(bitmap_header, 'BitmapOffset', size=4)
+        bitmap_header.create_chunk('BitmapOffset', size=4)
 
-        dib_header = self.client.create_chunk(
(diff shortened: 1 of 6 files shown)


python/veles/objects/blob.py
```diff
@@ -0,0 +1,28 @@
+from veles import network_pb2
+from veles.objects.base import LocalObject
+
+class Blob(LocalObject):
+    def __init__(self, client, proto_obj=None, id_path=None, parent=None):
+        super(Blob, self).__init__(client, proto_obj, id_path, parent)
+        self._client = client
```
MaciekPytel: this should be in base class

python/examples/parse_bmp.py
```diff
@@ -20,149 +20,119 @@ def parse_files(self):
...
-                '<i', file.data[self.client.cursor_value(file):
-                                self.client.cursor_value(file)+4])[0])
-            self.client.create_chunk(dib_header, 'Height', size=4)
-            self.client.create_chunk(dib_header, 'NumPlanes', size=2)
-            bits_per_pix = struct.unpack(
-                '<H', file.data[self.client.cursor_value(file):
-                                self.client.cursor_value(file)+2])[0]
-            self.client.create_chunk(dib_header, 'BitsPerPixel', size=2)
+            width_chunk = dib_header.create_chunk('Width', size=4)
+            width = struct.unpack('<i', width_chunk[0:4])[0]
```
altran01: Missing abs.

python/examples/parse_bmp.py
```diff
@@ -20,149 +20,119 @@ def parse_files(self):
...
+            height_chunk = dib_header.create_chunk('Height', size=4)
+            height = struct.unpack('<i', height_chunk[0:4])[0]
```
altran01: Here as well.

MaciekPytel: oops... Will fix.

python/examples/parse_bmp.py
```diff
@@ -20,149 +20,119 @@ def parse_files(self):
...
-                self.client.create_chunk(dib_header, 'XResolution', size=4)
-                self.client.create_chunk(dib_header, 'YResolution', size=4)
-                colors_used = struct.unpack(
-                    '<I',
-                    file.data[self.client.cursor_value(file):
-                              self.client.cursor_value(file)+4])[0]
-                self.client.create_chunk(dib_header, 'ColorsUsed', size=4)
-                self.client.create_chunk(dib_header, 'ColorsImportant', size=4)
+                compr_chunk = dib_header.create_chunk('Compression', size=4)
+                compression = struct.unpack('<i', compr_chunk[0:4])[0]
```
altran01: Compression is unsigned.

python/veles/objects/base.py
```diff
@@ -100,3 +104,18 @@ def __repr__(self):
...
+        if start is None:
+            start = self._cursor
+
+        if end is None and size is None:
+            raise ValueError(
+                'You need to specify either size or end of chunk')
+        if end is None:
+            end = start + size
+        if update_cursor:
+            self._cursor = end
```
altran01: Hmm now that I think that it is better if we update cursor after we create chunk - that way if something fails and we want to retry from IPython we still have the old cursor value.

MaciekPytel: Good point, will do.

python/veles/objects/chunk.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __iter__(self):
+        return self._blob[self._start:self._end].__iter__()
+
+    def __getitem__(self, arg):
+        if self._default_address_mode == self.AddressMode.ABSOLUTE:
+            return self._blob[arg]
+        if isinstance(arg, int):
+            return self._blob(arg + self._start)
+        if isinstance(arg, slice):
```
altran01: What do you think about adding possibility to use `chunk[:]` to get whole chunk content ? I think it is more convenient than `chunk[0:4]`

python/veles/veles_api.py
```diff
@@ -29,11 +30,18 @@ class ObjectTypes(object):
         SUB_BLOB = 2
         CHUNK = 3
 
+        CONSTRUCTORS = [
+            objects.LocalObject,  # ROOT
+            objects.Blob,  # FILE_BLOB
+            objects.Blob,  # SUB_BLOB
+            objects.Chunk  # CHUNK
```
altran01: I think dictionary type->class would be more appropriate, what do you think?

python/veles/objects/chunk.py
MaciekPytel: I don't see why not - will fix.

python/veles/veles_api.py
MaciekPytel: Fair enough, might as well not duplicate enums.

PR #34, opened 2017-01-30
Wip/ji/log database

jacek-izykowski:
Closed db view and log view can be reopened through "View" menu.

### DIFF
--- a/src/ui/veles_mainwindow.cc
+++ b/src/ui/veles_mainwindow.cc
@@ -1042,9 +1042,9 @@ void VelesMainWindow::init() {
   createLogWindow();
   createDb();
 
-  optionsDialog = new OptionsDialog(this);
+  options_dialog_ = new OptionsDialog(this);
 
-  connect(optionsDialog, &QDialog::accepted, [this]() {
+  connect(options_dialog_, &QDialog::accepted, [this]() {
     QList<QDockWidget*> dock_widgets = findChildren<QDockWidget*>();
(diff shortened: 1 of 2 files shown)


src/ui/veles_mainwindow.cc
```diff
@@ -984,6 +984,8 @@ bool MainWindowWithDetachableDockWidgets::splitDockWidgetImpl(
 VelesMainWindow::VelesMainWindow() : MainWindowWithDetachableDockWidgets() {
   setAcceptDrops(true);
   resize(1024, 768);
+  database_dock_widget_ = 0;
+  log_dock_widget_ = 0;
```
altran01: nit. we could use initializer lists instead

src/ui/veles_mainwindow.cc
```diff
@@ -1114,9 +1128,27 @@ void VelesMainWindow::updateParsers(dbif::PInfoReply reply) {
   }
 }
 
+void VelesMainWindow::showDatabase() {
+  if (database_dock_widget_ == nullptr) {
+    createDb();
+  }
+
+  database_dock_widget_->raise();
```
altran01: Hmm, one thing that might be slightly confusing about this implementation is that if database is in other main window (hidden under current one) user won't see any change after he click "Show database view" - although I don't know how hard is it to raise whole main windows in response.

src/ui/veles_mainwindow.cc

PR #35, opened 2017-01-30
Add basic handling of ChunkDataItem

altran01:


include/data/field.h
```diff
@@ -36,6 +36,7 @@ struct FieldHighType {
     STRING,
     POINTER,
     ENUM,
+    MODE_GUARD,
```
MaciekPytel: Maybe it would be better to always use the same name for guard? (hopefully) you always use enum with qualified name, so it's unambigous anyway.

Also it would be good for @koriakin to review adding those.

altran01: Simple enums won't allow it since the `GUARD` name wouldn't be unique (it would be possible with enum classes).

protobuf/network.proto
```diff
@@ -57,12 +85,8 @@ message Request {
...
     repeated uint64 id = 2;
 
-    string name = 101;
-    string comment = 102;
-
-    uint64 chunk_start = 201;
-    uint64 chunk_end = 202;
-    string chunk_type = 203;
+    LocalObject object = 101;
+    ChunkDataItem chunk_item = 102;
```
MaciekPytel: I don't particularly like how we include this in every request. Maybe we could have a RequestBody as a union based on Operation (which would be better named RequestType BTW).

python/veles/objects/chunk.py
```diff
@@ -15,6 +19,15 @@ def __init__(self, client, proto_obj=None, id_path=None, parent=None):
         self._end = proto_obj.chunk_end
         self._blob = parent.get_blob() if parent is not None else None
         self._cursor = self._start
+        self.items = [ChunkDataItem(item) for item in self._proto_obj.items]
+
+    def _from_another(self, other):
+        super(Chunk, self)._from_another(other)
+        self.items = other.items
+
+    def items_from_another(self, other):
```
MaciekPytel: Why is _from_another private and items_from_another not? Also s/items_from_another/copy_items? Finally shouldn't this method do self.items = other.items[:] at the very least? Or maybe even copy.deepcopy?

python/veles/objects/chunk.py
```diff
@@ -38,3 +51,45 @@ def __getitem__(self, arg):
...
+                          bit_width=8, high_pad=0, low_pad=0,
+                          mode=FieldHighMode.FIXED,
+                          shift=0,
+                          sign_mode=FieldSignMode.UNSIGNED,
+                          float_mode=FieldFloatMode.IEEE754_SINGLE,
+                          float_complex=False,
+                          string_mode=FieldStringMode.STRING_RAW,
+                          string_encoding=FieldStringEncoding.ENC_RAW,
+                          type_name=""):
+        # TODO big docstring
```
MaciekPytel: s/big/massive

python/veles/objects/chunk.py
```diff
@@ -38,3 +51,45 @@ def __getitem__(self, arg):
...
+        req.chunk_item.high_type.float_mode = float_mode
+        req.chunk_item.high_type.float_complex = float_complex
+        req.chunk_item.high_type.string_mode = string_mode
+        req.chunk_item.high_type.string_encoding = string_encoding
+        req.chunk_item.high_type.type_name = type_name
+        results = self._client._send_req(req)
+
+        self.items_from_another(results[0])
+        if update_cursor:
+            self._cursor = self.items[-1].end
```
MaciekPytel: What's the ordering of ChunkDataItems? Are they sorted by creation order? If not this may be assuming too much.

src/network/server.cc
```diff
@@ -140,6 +141,79 @@ void NetworkServer::getBlobData(PLocalObject target_object, QTcpSocket *client_c
...
 
+void NetworkServer::addChunkItem(PLocalObject blob, network::Response &resp,
+                                 PLocalObject object,
+                                 const network::ChunkDataItem &req_item) {
+  if (object->type() != dbif::CHUNK) {
+    resp.set_ok(false);
+    resp.set_error_msg("Items can only be added to chunk");
+  }
+  auto chunk = object.staticCast<ChunkObject>();
+  if (req_item.repack().endian() >= static_cast<int>(data::RepackEndian::GUARD)) {
```
MaciekPytel: I wonder if you can mess up some of this checks by supplying very large value on packet (to flip the int)?

python/veles/objects/chunk.py
altran01: Yes it should be private - I'll change it.
About name and copying - I created this method to "transfer" chunk items from object received from network to normal chunk, and in this use-case the network object isn't used afterwards. I can rename it to something like "_transfer_items" if you want (possibly also clearing items from other).

python/veles/objects/chunk.py
altran01: In C++ code they are pushed to back of vector - I'll make sure that it is never sorted.

src/network/server.cc
altran01: Hmm I think there isn't a possibility for error here - currently left side will be user controlled unsigned int and right side will be a small positive int, which shouldn't cause any problem - but I can be more explicit (use uint32_t in cast).

PR #37, opened 2017-02-17
pyc: Add a simple disassembler.

koriakin:


src/parser/unpyc.cc
```diff
@@ -203,18 +377,79 @@ void parseCode(dbif::ObjectHandle code) {
...
+    if (!found)
+      return;
+  }
+#if 0
+  for (auto &line : lines) {
+    // XXX set some sort of a prop with line no
+    bytecodeBlob->syncRunMethod<dbif::ChunkCreateRequest>(
+      QString("line_%1").arg(std::get<2>(line)), "pyc_line_tag", dbif::ObjectHandle(), std::get<0>(line), std::get<1>(line));
+  }
+#endif
```
altran01: What's the reason you decided to not include any information about line number in code blob ?

BTW. I didn't know github displays always false if-defs as comments - neat.

PR #39, opened 2017-02-20
HexEditWidget and NodeTreeWidget merged into single tab.

jacek-izykowski:
Features related to interactions between minimap and hex view
are out of the scope of this commit.


altran01:
Would it also be possible to remember which sidebars are open when moving hex-view tab around? Currently all tabs are automatically closed when you move tab in any way (apart from reordering tabs).

src/ui/hexedit.cc
```diff
@@ -413,6 +413,7 @@ void HexEdit::getRangeFromIndex(QModelIndex index, qint64 *start,
 void HexEdit::paintEvent(QPaintEvent *event) {
   QPainter painter(viewport());
 
+#if 0
```
altran01: At least add some comment as to why is this part of code permanently disabled?

src/ui/hexeditwidget.cc
```diff
@@ -137,8 +150,19 @@ void HexEditWidget::createActions() {
...
-  connect(show_node_tree_act_, SIGNAL(triggered()),
-      this, SLOT(showNodeTree()));
+  show_node_tree_act_->setCheckable(true);
+  show_node_tree_act_->setChecked(true);
+  connect(show_node_tree_act_, SIGNAL(toggled(bool)),
+      this, SIGNAL(showNodeTree(bool)));
+
+  show_minimap_act_ = new QAction(QIcon(":/images/show_minimap.png"),
+      tr("&Minimap"), this);
+  show_minimap_act_->setToolTip(tr("Node tree"));
```
altran01: copy-paste error :)

src/ui/veles_mainwindow.cc
```diff
@@ -297,6 +298,7 @@ void VelesMainWindow::createHexEditTab(QString fileName,
   QSharedPointer<QItemSelectionModel> selection_model(
       new QItemSelectionModel(data_model.data()));
 
+#if 0
```
altran01: Again some comment here ?

altran01:
Another bug I've noticed is that changing hex-edit columns number in options doesn't have an effect on already open tabs anymore - it is only applied to newly opened ones.

jacek-izykowski:
@altran01 Ok. But it seems to be out of the scope of this pull request.

src/ui/hexeditwidget.cc
```diff
@@ -265,5 +332,22 @@ void HexEditWidget::newBinData() {
   visualisation_act_->setEnabled(data_model_->binData().size() > 0);
 }
 
+void HexEditWidget::selectionChanged(qint64 start_addr,
+    qint64 selection_size) {
+  selection_label_->setText(
```
altran01: I think that one of assumptions we made about this label was that user should be able to easily select it and copy its content (this wasn't true for previous version as well, however we might want to change it now).

src/ui/hexedit.cc
```diff
@@ -434,6 +435,7 @@ void HexEdit::paintEvent(QPaintEvent *event) {
   painter.drawText(startMargin_ - startPosX_,
                    separatorLength - horizontalAreaSpaceWidth_,
                    statusBarText());
```
altran01: Overall I'm not a big fan of leaving unused code like this lying around - if someone would like to use it again later on then he could simply get it from git history. This introduces some unnecessary complications - like unused methods i.e. statusBarText - without checking that it now isn't called anywhere one could be surprised that editing it doesn't have any effect.

src/ui/hexedit.cc
```diff
@@ -434,6 +435,7 @@ void HexEdit::paintEvent(QPaintEvent *event) {
...
+#endif
 
   for (auto rowNum = startRow_;
```
altran01: After removing the code above, we now have an quite big empty area at bottom of our hex-view - maybe we should do something about it ?

src/ui/hexeditwidget.cc
jacek-izykowski: Fixed.

altran01:
Please rebase.

PR #41, opened 2017-02-21
Disassembler start.

koriakin:


python/veles/dis/reg.py
```diff
@@ -0,0 +1,79 @@
...
+    - name: a name displayed for this register, should be unique.
+    - width: width in bits.
+    """
+
+    def __init__(self, name, width):
+        self.name = name
+        self.width = width
+
+    def __str__(self):
+        return f"${self.reg.name}"
```
altran01: Unless I miss something BaseRegister doesn't have `reg` attribute ?

python/veles/dis/isa/falcon.py
```diff
@@ -0,0 +1,796 @@
...
+        IsaMatch(0xf, Insn("iord", arg_r3, arg_iorr)),
+    )
+
+
+    ### The parser.
+
+    # The default arguments in the following helper functions are a hack needed
+    # to use the instruction words inside the function - as class-level
+    # variables/fields, they are otherwise inaccessible until the class is
+    # constructed, and we need these functions to construct the parser.
```
altran01: WTF H4X :P

PR #44, opened 2017-02-22
Update README.rst

mkow:
Unify URLs to Veles homepage.


altran01:
You named it "unifying" URLs - but in other places (i.e. about view) we use codisec.com one :)

mkow:
Fixed.

PR #49, opened 2017-02-27
Wip/ji/shortcuts 211

jacek-izykowski:


altran01:
This allows one to use ctrl+f shortcut with more than one file open, but now it's extremely easy to use it in different file context than one would expect (and unfortunately the active dock changes that probably would help, aren't really working for me).

jacek-izykowski:
@altran01 Yes, we probably should merge this branch once wip/ji/active-dock is fixed and merged.

PR #57, opened 2017-03-06
Wip/mkow/new encoders

mkow:
Added new encoders: C data, C string, Text/Raw, Url


include/util/encoders/c_string_encoder.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace veles {
+namespace util {
+namespace encoders {
+
+class CStringEncoder : public IEncoder {
+ public:
+  QString encode(const QByteArray& data) override;
+  QString encodingDisplayName() override;
+  QByteArray decode(const QString& str) {return {};}
```
altran01: Why is this needed in encoder?

src/util/encoders/hex_encoder.cc
```diff
@@ -14,35 +14,29 @@
...
+  return "Hex";
+}
+
+QString HexEncoder::encode(const QByteArray& data) {
+  QByteArray ba = data.toHex();
+  return QString::fromLatin1(ba);
+}
+
+QByteArray HexEncoder::decode(const QString& str) {
+  return QByteArray().fromHex(str.toLatin1());
```
altran01: `fromHex` is static method, so there is no need to create empty `QByteArray` object.

altran01:
One thing I'm not sure about - whether should we copy anything to clipboard when our selection is 0 bytes (in case of C-data it is even code that doesn't compile :P).

mkow:
Hmm, good point, but I think we can leave it as is, because after we reimplement cursor in hex view there will always be something to copy (the only edge case is for empty file, which is quite rare).

Also: IMO it's still better to copy an empty C string/C array (even if it doesn't compile), we shouldn't have a function which does nothing in some cases (i.e.: no effects and no errors at the same time).

PR #67, opened 2017-03-16
Convert dis classes to support msgpack serialization

altran01:
Also expand msgpack wrapper code to correctly handle those situations


python/delete
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 
 import sys
-from veles.scli import client
 
-from veles.common import base
+from veles.schema import nodeid
+from veles.scli import client
```
koriakin: nożyczkami go na 2 PRy.

altran01: Done.

PR #69, opened 2017-03-16
Expand msgpack wrapper code

altran01:
This is required to handle more complicated situations like dis code


python/veles/schema/nodeid.py
```diff
@@ -0,0 +1,54 @@
...
+        return self._bytes
+
+    def to_bytes(self):
+        return self.bytes
+
+    def __str__(self):
+        return binascii.b2a_hex(self.bytes).decode('ascii')
+
+    def __repr__(self):
+        return 'ObjectID("{}")'.format(str(self))
```
koriakin: error

python/veles/asrv/proto.py
```diff
@@ -96,7 +98,10 @@ def data_received(self, data):
         self.unpacker.feed(data)
         while True:
             try:
-                msg = definitions.MsgpackMsg.load(self.unpacker)
+                msg = definitions.MsgpackMsg.load(self.unpacker.unpack())
+                if not isinstance(msg, definitions.MsgpackMsg):
```
koriakin: This shouldn't be needed.

python/veles/messages/definitions.py
```diff
@@ -12,45 +12,17 @@
...
-            obj[field.name] = field.__get__(self)
-        return packer.pack(obj)
-
-    def __str__(self):
-        msg = '{}: {}'.format(self.msg_type, self.to_dict())
-        return msg
-
-    __repr__ = __str__
+class MsgpackMsg(model.PolymorphicModel):
+    object_type = None
```
koriakin: object_type shouldn't be needed

koriakin:
It would be good to split off the ObjectID -> NodeID rename - that can be merged immediately.

python/veles/messages/definitions.py
altran01: It wasn't.

PR #70, opened 2017-03-16
Various code fixes

mkow:
Changes:
- Captions in trigram view are now disabled by default (not too useful and quite ugly)
- Many small code fixes


include/util/math.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef VELES_UTIL_MATH_H
+#define VELES_UTIL_MATH_H
+
+#include <type_traits>
+
+namespace veles {
+namespace util {
+namespace math {
+
+template<typename T1, typename T2>
```
koriakin: Napisz tylko, że do kasacji jak będziemy mogli użyć C++17

src/data/repack.cc
```diff
@@ -15,20 +15,16 @@
  *
  */
 #include "data/repack.h"
+
 #include <stdlib.h>
 
+#include "util/math.h"
```
altran01: Do we have any convention about include sorting? In previous commit you moved includes in a way that doesn't match what you do here.

mkow: I try to format the according to Google C++ Style Guide: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
Which commit exactly has different order?

include/util/math.h
mkow: Fixed.

src/data/repack.cc
altran01: Mhm, then never mind - I was confused by imports from our project both on top and bottom of list, but it seems to be in line with those guidelines.

altran01:
About the British -> American change in visualisation, I'm not very big supporter of this part of the change - but if you insist why don't change other words British -> American?

mkow:
@altran01
IMO we should stick to American English because it seems to be the standard in programming. If you know other places where we have British English I can change them too, better do it now, when we have relatively small codebase.

Also, for the -ize-like endings, it seems that even in British English "z" is more common/recommended choice: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-ise.2C_-ize_.28-isation.2C_-ization.29

altran01:
Two examples I can recall:
- initialis*
- normalis*

mkow:
Done.

PR #71, opened 2017-03-16
Add an interface to async connections.

koriakin:
Subsequent PRs will make asrv use this interface.

### DIFF
--- a/python/veles/async_conn/conn.py
+++ b/python/veles/async_conn/conn.py
@@ -1,0 +1,69 @@
+# Copyright 2017 CodiLime
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


python/veles/async_conn/conn.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Creates a node with the given data.  If ``id`` is None, a random id
+        will be generated.  Returns a tuple of (node, awaitable of node) -
+        the node is available immediately (and can be linked to other objects,
+        or used as a parent), and the awaitable will be resolved once
+        the operation completes on the server.
+        """
+        if id is None:
+            id = ObjectID()
+        obj = self.get_node_norefresh(id)
```
altran01: Will `get_node_norefresh` be responsible for creating random id? `ObjectID()` doesn't generate one - it creates one with 24 zero bytes, and it won't be very useful for linking.

python/veles/async_conn/subscriber.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    A subscriber of node modifications.  To watch for node modifications,
+    create a subclass of this, override object_changed and error callbacks,
+    and create an instance.  When the subscription is no longer used, call
+    cancel().
+    """
+
+    def __init__(self, obj):
+        self.obj = obj
+        self.live = True
```
altran01: nit. alive?

koriakin: Done.

PR #75, opened 2017-03-17
int_to_bytes: Ensure bytes are actually returned.

koriakin:


python/veles/tests/compatibility/test_int_bytes.py
```diff
@@ -34,6 +34,7 @@ def test_int_from_bytes(self):
         self.assertEqual(int_from_bytes(x, 'big'), z)
 
     def test_int_to_bytes(self):
+        self.assertTrue(isinstance(int_to_bytes(0, 0, 'little'), bytes))
```
altran01: assertIsInstance is a thing IIRC.

koriakin: fixed

PR #78, opened 2017-03-18
db: Fix & test bindata support.

koriakin:


python/veles/db.py
```diff
@@ -292,6 +295,13 @@ def set_data(self, obj_id, key, data):
     def get_bindata(self, obj_id, key, start=0, end=None):
         if not isinstance(obj_id, NodeID):
             raise TypeError('node id has wrong type')
+        start = operator.index(start)
+        if end is not None:
+            end = operator.index(end)
+        if start < 0:
+            raise ValueError('start must not be negative')
+        if end is not None and end < start:
```
koriakin: FWIW this is intended, it'll be explained in the docstring

python/veles/db.py
```diff
@@ -292,6 +295,13 @@ def set_data(self, obj_id, key, data):
     def get_bindata(self, obj_id, key, start=0, end=None):
```
koriakin: Yep, docstrings are in the TODO list...

PR #79, opened 2017-03-18
Order-preserving bigint encoding.

koriakin:
This encoding allows us to store bigints in the database (pos_start
and pos_end columns) and have them correctly sorted.

Also, while we're at it, let's replace the current msgpack encoding
scheme, since it's quite redundant.

### DIFF
--- a/python/veles/db.py
+++ b/python/veles/db.py
@@ -20,14 +20,14 @@ from veles.proto import msgpackwrap
 from veles.schema.nodeid import NodeID
 from veles.proto.node import Node
 from veles.compatibility.int_bytes import int_from_bytes
+from veles.util.bigint import bigint_encode, bigint_decode
 
 
-APP_ID = int_from_bytes(b'ml0n', 'big')
+APP_ID = int_from_bytes(b'mln0', 'big')
 
 
 # TODO:
 #
-# - bignum positions
 # - use root id
 # - link support
 # - xref support
@@ -37,6 +37,14 @@ APP_ID = int_from_bytes(b'ml0n', 'big')
 if six.PY3:
     def buffer(x):
         return x
+
+
+def db_bigint_encode(val):
+    return None if val is None else buffer(bigint_encode(val))
+
+
+def db_bigint_decode(val):
+    return None if val is None else bigint_decode(bytes(val))
 
 
 class Database:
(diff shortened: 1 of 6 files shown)


python/veles/db.py
```diff
@@ -20,14 +20,14 @@
 from veles.schema.nodeid import NodeID
 from veles.proto.node import Node
 from veles.compatibility.int_bytes import int_from_bytes
+from veles.util.bigint import bigint_encode, bigint_decode
 
 
-APP_ID = int_from_bytes(b'ml0n', 'big')
+APP_ID = int_from_bytes(b'mln0', 'big')
```
altran01: eh :P

python/veles/util/bigint.py
```diff
@@ -0,0 +1,72 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import operator
+
+from veles.compatibility.int_bytes import int_to_bytes, int_from_bytes
+
+
+HDR_BIAS = 0x80000000
```
altran01: This seems like a little more than I hope we will ever need to use, but on the other hand we probably won't care that much about some additional bytes used by our bignums in DB and/or over network.

python/veles/util/bigint.py
```diff
@@ -0,0 +1,72 @@
...
+    invalid encodings are detected.
+    """
+    if not isinstance(val, bytes):
+        raise TypeError('encoded bigint must be bytes')
+    if len(val) < HDR_LEN:
+        raise ValueError('encoded bigint too short')
+    hdr = int_from_bytes(val[:HDR_LEN], 'big')
+    sz = abs(hdr - HDR_BIAS)
+    rest = val[HDR_LEN:]
+    rval = int_from_bytes(rest, 'big')
```
altran01: Move after validation below.

python/veles/util/bigint.py
koriakin: Done.

python/veles/util/bigint.py
koriakin: Done.

python/veles/util/bigint.py
koriakin: Yeah, 16-bit should be enough, but who cares about 2 more bytes...

PR #81, opened 2017-03-18
Schema cleanups & improvements

koriakin:
Changes:

- schema validation errors raise SchemaError.
- setting value of a field and deserializing it are now distinct operations
  and are more closely validated - it is now illegal to assign a dict to
  a model field, for example.
- Extension replaced with new NodeID and BinData fields.
- Array split into List and Set, which now only accept list/set, respectively.
- the type parameters of List/Set/Map now come before the optional parameter,
  and are intended to be used as positional params.
- a default value may now be set for a field.
- List/Set/Map have a default value of []/set()/{} by default.
- a new UnsignedInteger field is introduced (it's just an Integer with
  minimum=0)
- SmallInteger and SmallUnsignedInteger fields are introduced - these
  are limitted to int64_t and uint64_t range, respectively, and will
  be represented as such in C++.
- the dump method of model no longer takes a packer, and returns a dict
  instead.

### DIFF
--- a/python/veles/proto/messages.py
+++ b/python/veles/proto/messages.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from veles.proto import node
-from veles.schema import nodeid, model, fields
+from veles.proto.node import Node
+from veles.schema import model, fields
 
 
 class MsgpackMsg(model.PolymorphicModel):
@@ -23,7 +23,7 @@ class MsgpackMsg(model.PolymorphicModel):
 class MsgConnect(MsgpackMsg):
     object_type = 'connect'
 
-    proto_version = fields.Integer(minimum=1)
+    proto_version = fields.SmallInteger(minimum=1)
     client_name = fields.String(optional=True)
     client_version = fields.String(optional=True)
     client_description = fields.String(optional=True)
@@ -34,7 +34,7 @@ class MsgConnect(MsgpackMsg):
 class MsgConnected(MsgpackMsg):
     object_type = 'connected'
 
-    proto_version = fields.Integer(minimum=1)
+    proto_version = fields.SmallInteger(minimum=1)
     server_name = fields.String()
     server_version = fields.String()
 
@@ -50,7 +50,7 @@ class MsgProtoError(MsgpackMsg):
     object_type = 'proto_error'
 
     code = fields.Integer()
-    msg = fields.String(optional=True)
+    msg = fields.String()
 
 
 class MsgRegisterMethod(MsgpackMsg):
@@ -64,11 +64,10 @@ class MsgRegisterTrigger(MsgpackMsg):
 class MsgList(MsgpackMsg):
     object_type = 'list'
 
-    parent = fields.Extension(obj_type=nodeid.NodeID, optional=True)
-    tags = fields.Array(elements_types=[fields.Map(
-        keys_types=[fields.String()], values_types=[fields.Boolean()])])
-    qid = fields.Integer()
-    sub = fields.Boolean()
+    parent = fields.NodeID(optional=True)
+    tags = fields.Set(fields.String())
+    qid = fields.SmallUnsignedInteger()
+    sub = fields.Boolean(default=False)
     pos_start = fields.Integer(optional=True)
     pos_end = fields.Integer(optional=True)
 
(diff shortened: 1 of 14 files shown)


python/veles/proto/messages.py
```diff
@@ -64,71 +64,68 @@ class MsgRegisterTrigger(MsgpackMsg):
 class MsgList(MsgpackMsg):
     object_type = 'list'
 
-    parent = fields.Extension(obj_type=nodeid.NodeID, optional=True)
-    tags = fields.Array(elements_types=[fields.Map(
-        keys_types=[fields.String()], values_types=[fields.Boolean()])])
-    qid = fields.Integer()
-    sub = fields.Boolean()
+    parent = fields.NodeID(optional=True)
+    tags = fields.Set(fields.String())
```
altran01: This change breaks `listsub`.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
+        if not isinstance(element, Field):
+            raise TypeError("List element must be a Field.")
+        self.element = element
+
+    def __set_name__(self, owner, name):
+        super(List, self).__set_name__(owner, name)
+        self.element.__set_name__(owner, name + '.element')
+
+    def _validate(self, value):
+        if not isinstance(value, list):
```
koriakin: Good idea.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
+            raise SchemaError(
+                'Attribute {} has to be a list'.format(self.name))
+        for val in value:
+            self.element.validate(val)
+
+    def _load(self, value):
+        if not isinstance(value, list):
```
koriakin: Won't work here - the items of the list are serialized, they likely won't pass validation for unserialized values.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
 from veles.compatibility import pep487
-from . import model
+from veles.proto.exceptions import SchemaError
+from veles.data import bindata
+from . import nodeid
 
 
 class Field(pep487.NewObject):
-    def __init__(self, optional=False):
+    def __init__(self, optional=False, default=None):
```
altran01: Sounds good.

python/veles/schema/model.py
```diff
@@ -12,47 +12,66 @@
...
     def load(cls, val):
-        return cls.from_dict(val)
-
-    def dump(self, packer):
-        return packer.pack(self.to_dict())
+        if not isinstance(val, dict):
+            raise SchemaError('a serialized model must be a dict')
+        val = dict(val)
+        res = cls.__new__(cls)
+        for field in cls.fields:
```
koriakin: The idea, like for pickle, is to have this work even if __init__ takes different arguments.  But I'm not really convinced either way...

python/veles/schema/fields.py
altran01: Right - I forgot that it would call `_validate` from `List`. I wanted to use it to also check `optional`, but overall I think we can get rid of `optional` in `List`, `Set` and `Map` altogether since they aren't used anywhere for those types.
Also we can still use base `_validate` here.

python/veles/schema/fields.py
koriakin: Done.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
+        ]
+
+    def _dump(self, value):
+        return [
+            self.element.dump(val)
+            for val in value
+        ]
+
+
+class Set(Field):
```
koriakin: Done.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
+        }
+
+    def _dump(self, value):
+        return [
+            self.element.dump(val)
+            for val in value
+        ]
 
 
 class Map(Field):
```
koriakin: Done.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
-        if isinstance(value, dict):
-            value = self.local_type.load(value)
+class Object(Field):
+    def __init__(self, value_type, optional=False, default=None):
+        super(Object, self).__init__(optional, default)
+        self.value_type = value_type
 
-        if isinstance(value, self.local_type):
-            return value
+    def _load(self, value):
```
koriakin: Fixed.

python/veles/schema/fields.py
```diff
@@ -15,214 +15,281 @@
...
+        if value is None:
+            self.validate(None)
+            return None
+        return self.value_type.load(value)
 
-        raise ValueError(
-            'Attribute {} has to be {} type or dict that maps to it.'.format(
-                self.name, self.local_type))
+    def _dump(self, value):
```
koriakin: Fixed.

python/veles/schema/fields.py
koriakin: Done.

python/veles/proto/messages.py
koriakin: Fixed.

PR #82, opened 2017-03-19
Add a PosFilter model.

koriakin:


python/veles/proto/node.py
```diff
@@ -24,3 +24,51 @@ class Node(model.Model):
...
+        Constructs a filter that returns chunks overlapping with a given
+        range of positions (left-inclusive, right-exclusive, like pos_start
+        and pos_end).
+        """
+        return cls(
+            start_to=pos_to-1,
+            end_from=pos_from+1,
+        )
+
+    def matches(self, node):
```
altran01: I assume this will be used in of remaining PRs or near future?

koriakin: It's used in the new list subscriber implementation.

python/veles/proto/node.py
```diff
@@ -24,3 +24,51 @@ class Node(model.Model):
...
+    of None is treated as -inf, and pos_end of None is treated as +inf.
+    """
+
+    start_from = fields.Integer(optional=True)
+    start_to = fields.Integer(optional=True)
+    end_from = fields.Integer(optional=True)
+    end_to = fields.Integer(optional=True)
+
+    @classmethod
+    def intersects_with(cls, pos_from, pos_to):
```
mkow: `intersecting_with` seems to be a better name for this.

koriakin: Fixed.

PR #84, opened 2017-03-19
db: Use NodeID.root_id

koriakin:


altran01:
I don't see anything wrong here, but I'd like to be able to use correctly working listsub to test it "manually".

koriakin:
Rebased - listsub should be working now.

PR #85, opened 2017-03-19
schema.fields: Always call super().__init__ last.

koriakin:
Without this, validating a default value will fail.


koriakin:
PR nie do review, pójdzie squash do #81

PR #86, opened 2017-03-19
Use PosFilter everywhere.

koriakin:


python/veles/asrv/srv.py
```diff
@@ -49,21 +49,11 @@ def kill(self):
...
-        for tags in self.tags:
-            for k, v in tags:
-                if v != (k in obj.tags):
-                    break
-            else:
-                return True
-        return False
+        if not self.pos.matches(obj):
+            return False
+        if not self.tags <= obj.tags:
```
altran01: Mhm, so this fixes listers, so now we remove option to filter by objects that don't have some tag, right?

koriakin: Yep.  YAGNI, IMO.

koriakin:
Squashed into #82

PR #87, opened 2017-03-19
asrv: Rename files

koriakin:


python/srv
```diff
@@ -17,26 +17,27 @@
...
 if host == 'UNIX':
     print('Starting UNIX server.')
-    loop.run_until_complete(veles.asrv.proto.unix_server(srv, port))
+    loop.run_until_complete(unix_server(conn, port))
 else:
     print('Starting TCP server.')
-    loop.run_until_complete(veles.asrv.proto.tcp_server(srv, host, int(port)))
+    loop.run_until_complete(tcp_server(conn, host, int(port)))
 print('Ready.')
 # loop.add_signal_handler(signal.SIGINT, loop.stop)
```
mkow: Not related to this PR, but why is this commented out? Please write a comment with reason or delete.

python/veles/server/proto.py
```diff
@@ -266,11 +266,9 @@ def send_obj_gone(self, qid):
         ))
 
 
-@asyncio.coroutine
-def unix_server(srv, path):
-    return (yield from srv.loop.create_unix_server(lambda: Proto(srv), path))
+async def unix_server(conn, path):
```
mkow: nit: rename to create_unix_server

python/veles/server/proto.py
```diff
@@ -266,11 +266,9 @@ def send_obj_gone(self, qid):
...
+    return await conn.loop.create_unix_server(lambda: ServerProto(conn), path)
 
 
-@asyncio.coroutine
-def tcp_server(srv, ip, port):
-    return (yield from srv.loop.create_server(lambda: Proto(srv), ip, port))
+async def tcp_server(conn, ip, port):
```
mkow: nit: rename to create_tcp_server

python/srv
koriakin: @altran01

altran01: I don't see any near future where this can be restored - the problem is that `add_signal_handler` isn't available on Windows so we can't use it :(
So lets remove it.

koriakin: try: ... except AttributeError: pass?

python/veles/server/proto.py
koriakin: Done

python/veles/server/proto.py
koriakin: Done

python/srv
koriakin: Done

PR #93, opened 2017-03-22
wip/ji/conn-dialog

jacek-izykowski:


src/ui/connectiondialog.cc
```diff
@@ -0,0 +1,174 @@
...
+
+  unsigned long random_key = random_int();
+  ui_->key_line_edit->setText(QString("%1").arg(random_key, 16, 16, QChar('0')));
+}
+
+void ConnectionDialog::userAsClientName() {
+  QString client_name;
+#if defined(Q_OS_LINUX) || defined(Q_OS_MAC)
+  client_name = qgetenv("USER");
+#else // Windows
```
mkow: Nit: Change this to #elif Q_OS_WINDOWS and add #error in #else.

src/ui/connectionmanager.cc
```diff
@@ -0,0 +1,181 @@
...
+      << QString("%1:%2").arg(connection_dialog_->serverHost())
+      .arg(connection_dialog_->serverPort());
+
+  QTextStream out(veles::ui::LogWidget::output());
+  out
+      << "Trying to start a new server..." << endl
+      << "    working directory: " << directory_path << endl
+      << "    python script name: " << server_file_name << endl
+      << "    arguments:" << endl;
+  for (auto arg : arguments) {
```
mkow: `const auto& arg`

src/ui/connectiondialog.cc
```diff
@@ -0,0 +1,174 @@
...
+
+void ConnectionDialog::newServerToggled(bool toggled) {
+  ui_->database_label->setEnabled(toggled);
+  ui_->database_line_edit->setEnabled(toggled);
+  ui_->select_database_button->setEnabled(toggled);
+
+  ui_->server_executable_label->setEnabled(toggled);
+  ui_->server_executable_line_edit->setEnabled(toggled);
+  ui_->select_server_executable_button->setEnabled(toggled);
+}
```
mkow: We should also disable "random" button for key generation when you're connecting to an existing server (it doesn't make any sense then).

src/ui/connectiondialog.cc
```diff
@@ -0,0 +1,174 @@
...
+  db_file_dialog_->setNameFilters(file_name_filters);
+
+  connect(ui_->select_database_button, &QPushButton::clicked,
+        db_file_dialog_, &QFileDialog::show);
+  connect(db_file_dialog_, &QFileDialog::fileSelected,
+      this, &ConnectionDialog::databaseFileSelected);
+
+  server_file_dialog_ = new QFileDialog(this);
+  server_file_dialog_->setAcceptMode(QFileDialog::AcceptOpen);
+  server_file_dialog_->setFileMode(QFileDialog::ExistingFile);
```
mkow: IMO this should be filtered using `Python script (*.py)`.

koriakin: Currently, srv is an extensionless executable file.  I suppose we could change it...

altran01:
Eh it's hard to respond to general comment on github /rant

> IMO we should refactor File->Options dialog now, because it may confuse user (especially the option "Enable network server"). It should say something like "Enable API server", but feel free to pick a better name for this.

As far as I understand if we move to our new server architecture we will want to "fix" the previous network API and protobuf.

src/ui/connectionmanager.cc
```diff
@@ -0,0 +1,190 @@
...
+  server_process_->setProcessEnvironment(env);
+
+  QStringList arguments;
+  arguments
+      << server_file_name
+      << connection_dialog_->databaseFile()
+      << QString("%1:%2").arg(connection_dialog_->serverHost())
+      .arg(connection_dialog_->serverPort());
+
+  QString python_interpreter_executable("python3");
```
altran01: By default when installing Python on Windows it doesn't create python2/3 executables - on this platform you should instead use py.exe which checks in Windows registry where interpreter for requested Python version is located.

However I'm thinking about how we should handle starting the server in the long run - current approach launches it in global Python env which requires user to install our requirements in global Python env - which most likely is not desirable.

src/ui/veles_mainwindow.cc
```diff
@@ -116,6 +117,15 @@ void VelesMainWindow::init() {
       }
     }
   });
+
+  tool_bar_ = addToolBar("Connection");
+  connection_notification_widget_ = new ConnectionNotificationWidget(this);
```
altran01: Disable right click on this widget (it allows you to hide it permanently).

src/ui/connectiondialog.cc
```diff
@@ -0,0 +1,176 @@
...
+  };
+  ui_->key_line_edit->setText(gen_key_part() + gen_key_part() + gen_key_part()
+                              + gen_key_part());
+}
+
+void ConnectionDialog::userAsClientName() {
+  QString client_name;
+#if defined(Q_OS_LINUX) || defined(Q_OS_MAC)
+  client_name = qgetenv("USER");
+#elif defined(Q_OS_WINDOWS)
```
jacek-izykowski: Yes, it's Q_OS_WIN actually. Fixed.

src/ui/connectionmanager.cc
```diff
@@ -0,0 +1,190 @@
...
+  return kill_locally_created_server_action_;
+}
+
+void ConnectionManager::locallyCreatedServerStarted() {
+  QTextStream out(veles::ui::LogWidget::output());
+  out << "Process of locally created server started." << endl;
+  kill_locally_created_server_action_->setEnabled(true);
+}
+
+void ConnectionManager::locallyCreatedServerFinished(int exit_code,
```
jacek-izykowski: Good point. Done.

src/ui/connectionmanager.cc
jacek-izykowski: Ok. For now I'm going to leave it with ifdefs to pick correct python executable.

src/ui/veles_mainwindow.cc
jacek-izykowski: Done.

altran01:
I'm not sure where lies the problem and is it only Windows specific problem, but when I try to start the server and it fails for whatever reason I can't see full process output if it crashes (including the most interesting part - stacktrace of exception that caused process to crash).

src/ui/connectionmanager.cc
```diff
@@ -0,0 +1,205 @@
...
+  // Can be done once C++ client library is created.
+}
+
+void ConnectionManager::serverProcessReadyRead() {
+  char buf[10240];
+  if(server_process_ && (server_process_->canReadLine()
+      || server_process_->bytesAvailable() > (qint64)sizeof(buf))) {
+    qint64 len = server_process_->readLine(buf, sizeof(buf));
+    if(len > 0) {
+      LogWidget::output()->write(buf, len);
```
adamiwaniuk: when output contains non ASCII characters (for exmaple **Ś**witezianka) then it doesn't display corretly on windows. Also there is some buffering issues. For example output from script "Ready" is not displayed until there is more characters in output buffer

src/ui/connectiondialog.ui
```diff
@@ -0,0 +1,341 @@
...
+          </size>
+         </property>
+         <property name="minimum">
+          <number>1</number>
+         </property>
+         <property name="maximum">
+          <number>65535</number>
+         </property>
+         <property name="value">
+          <number>1905</number>
```
adamiwaniuk: why 1905 ?

jacek-izykowski: Without very specific reason. Why not? Do you have any suggestions?

adamiwaniuk: We have this discussion with Altran, and my proposal was 3135, because V3135 -> VELES :p

jacek-izykowski: Changed.

src/ui/connectionmanager.cc
jacek-izykowski: Fixed.

PR #96, opened 2017-03-22
db: BinData paging.

koriakin:


koriakin:
In theory, it should already be covered by the existing bindata tests - that's why I used ridiculously long bindata in these in the first place.

PR #97, opened 2017-03-23
Wip/mwk/exception field

koriakin:


python/veles/schema/fields.py
```diff
@@ -283,3 +283,24 @@ def _load(self, value):
 
     def _dump(self, value):
         return six.text_type(value.name)
+
+
+class Exception(Field):
```
altran01: Why did you do it like this instead of `Object` with `value_type` set to `VelesException` and make `VelesException` be a model (albeit a very custom one)? Then you wouldn't need to do all this additional validation here.

koriakin: Done-ish

koriakin:
Note: currently breaks the build, since VelesException has no support for the C++ generator.  To be corrected...

PR #99, opened 2017-03-24
Minimap doesn't pop up during visualisation initialization

jacek-izykowski:


jacek-izykowski:
@altran01 I've done the same thing for hex view.

altran01:
As discussed in person now it crashes on Windows on tab closing.

PR #100, opened 2017-03-24
wip/ji/ui-server-shutdown

jacek-izykowski:


altran01:
I know I'm a little late to the party, but I have a few comments to this PR.
This approach to killing server on exiting doesn't remove ugly error message when closing Veles - on Windows we removed console so user can't see it, but on Ubuntu he will see:
`QProcess: Destroyed while process ("python3") is still running.`
Another problem is that on Windows the "Kill locally created server" doesn't seem to work (it works fine on Ubuntu) - I click it multiple times but created process doesn't stop and still provides output to log widget.

altran01:
Another most likely undesired behaviour is starting another local server when one is already running - it results in no process running in the end and outputs following error to Veles process output:
`QCoreApplication::postEvent: Unexpected null receiver`

PR #104, opened 2017-03-27
asrv: transaction support

koriakin:


python/veles/server/conn.py
```diff
@@ -66,6 +77,230 @@ def get_node_norefresh(self, obj_id):
...
+    def _check_ok_list(self, el):
+        parent = self.get_node_norefresh(el.parent)
+        if parent.node is None and parent != self.root:
+            return False
+        nodes = self.db.list(el.parent, el.tags, el.pos_filter)
+        return el.nodes == nodes
+
+    def _checks_ok(self, checks):
+        for ch in checks:
+            handlers = {
```
altran01: nit. it probably won't have any noticeable effect on performance, but it hurt me when I saw this dict constructed inside loop :P

python/veles/server/conn.py
```diff
@@ -66,6 +77,230 @@ def get_node_norefresh(self, obj_id):
...
+
+    def transaction(self, checks, ops):
+        if not self._checks_ok(checks):
+            return bad_future(PreconditionFailedError())
+        try:
+            with Transaction(self) as xact:
+                for op in ops:
+                    node = self.get_node_norefresh(op.node)
+                    xact.save(node)
+                    handlers = {
```
altran01: nit. same comment about dict in loop.

python/veles/server/conn.py
```diff
@@ -66,6 +77,230 @@ def get_node_norefresh(self, obj_id):
...
+        if op.truncate:
+            new_len = op.start + len(op.data)
+            change_end = max(old_len, new_len)
+        else:
+            new_len = max(op.start + len(op.data), old_len)
+            change_end = op.start + len(op.data)
+        for sub in node.bindata_subs.get(op.key, ()):
+            if sub.end is not None and sub.end <= op.start:
+                continue
+            if change_end <= sub.start:
```
altran01: This condition is different from the one in `_send_subs_bindata` - is this intended?

python/veles/db.py
```diff
@@ -448,9 +455,10 @@ def set_bindata(self, id, key, start, data, truncate=False):
         ])
 
         # We're done here.
-        self.db.commit()
+        if commit:
+            self.commit()
 
-    def delete(self, id):
+    def delete(self, id, commit=True):
```
altran01: Hmm, maybe we can also use it in `AsyncLocalNode.delete` to only commit one delete instead of commit for every child of deleted Node?

python/veles/server/conn.py
koriakin: Yes - change_end is computed exactly in this function, and cannot be None (the non-transaction path is sloppier).

python/veles/db.py
koriakin: We could, but this path is going to be fixed soon anyway.

python/veles/server/conn.py
koriakin: Done.

python/veles/server/conn.py
koriakin: Done.

PR #107, opened 2017-03-29
ChunkDataItem model.

koriakin:


python/veles/proto/chunk.py
```diff
@@ -0,0 +1,97 @@
...
+    pos_end = fields.Integer()
+    ref = fields.NodeID()
+
+
+class ChunkDataItemSubblob(ChunkDataItem):
+    name = fields.String()
+    ref = fields.NodeID()
+
+
+class ChunkDataItemField(ChunkDataItem):
```
altran01: In cpp code ChunkDataItemField also had vector of refs - was removing it intended (since other chunk-items still have refs).

koriakin: AFAIK it's not used anywhere, so let's skip it for now.

PR #109, opened 2017-03-29
Tracer

koriakin:


python/veles/async_conn/tracer.py
```diff
@@ -0,0 +1,196 @@
...
+
+        async def inner():
+            try:
+                node = (await anode).node
+            except ObjectGoneError:
+                self.checks.append(check.CheckGone(
+                    node=id,
+                ))
+                raise
+            else:
```
altran01: nit. not needed since you re-raise exception.

PR #111, opened 2017-03-30
fixed support for python2 in client

adamiwaniuk:


python/veles/schema/fields.py
```diff
@@ -191,6 +191,17 @@ def cpp_type(self):
...
 
+    def __set__(self, instance, value):
+        if isinstance(value, str):
+            value = six.text_type(value)
+        super(String, self).__set__(instance, value)
+
+    def _dump(self, value):
+        return six.text_type(value)
+
+    def _validate(self, value):
```
koriakin: No.

The main problem with Python 2 is that unicode and str sort of morph into each other and you can never quite tell if you're dealing with one or the other.  This class is specifically designed to reject str on Python 2, to make sure we're always dealing with unicode.

Your changes make it possible to create a model instance with a String field set to either str or unicode (by unpacking it from msgpack), thrashing the type guarantees that are the whole point of the Model system.  This is caught by the unittests, which are (predictably) broken.

python/veles/schema/model.py
```diff
@@ -246,6 +246,7 @@ def __init_subclass__(cls, **kwargs):
         if not hasattr(cls, 'object_types'):
             cls.object_types = {}
         elif cls.object_type is not None:
+            cls.object_type = six.text_type(cls.object_type)
```
koriakin: not a bad idea, but needs verification that object_type is of a sane type (since every object is str'able in python)

python/veles/schema/fields.py
koriakin: Also, `from __future__ import unicode_literals`

PR #119, opened 2017-04-01
Query support.

koriakin:


python/veles/server/node.py
```diff
@@ -209,13 +255,13 @@ def delete(self):
...
+            for key, subs in list(self.data_subs.items()):
+                for sub in list(subs):
                     sub.error(ObjectGoneError())
-            for key, subs in self.bindata_subs.items():
-                for sub in subs:
+            for key, subs in list(self.bindata_subs.items()):
+                for sub in list(subs):
                     sub.error(ObjectGoneError())
-            for sub in self.list_subs:
+            for sub in list(self.list_subs):
```
koriakin: It protects against subscriber callbacks modifying the subscriber set during iteration.

This will be more properly fixed by nuking all non-transaction modification paths (transactions pre-calc a list of affected subscribers).

PR #126, opened 2017-04-05
Triggers - plugin side.

koriakin:
This just adds registration of trigger handlers - they're not called yet.

### DIFF
--- a/python/veles/server/proto.py
+++ b/python/veles/server/proto.py
@@ -35,6 +35,7 @@ from veles.async_conn.plugin import (
     MethodHandler,
     QueryHandler,
     BroadcastHandler,
+    TriggerHandler,
 )
 from veles.proto.exceptions import (
     VelesException,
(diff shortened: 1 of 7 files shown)


python/veles/server/proto.py
```diff
@@ -600,6 +630,29 @@ def send_msg(self, msg):
...
         del self.pbids[msg.pbid]
 
+    async def msg_plugin_trigger_register(self, msg):
+        if msg.phid in self.phids:
+            raise SubscriptionInUseError()
+        handler = RemoteTriggerHandler(msg.name, msg.tags, self, msg.phid)
+        self.phids[msg.phid] = handler
+        self.conn.register_trigger_handler(handler)
+
+    async def msg_plugin_trigger_gone(self, msg):
```
altran01: s/msg_plugin_trigger_gone/msg_plugin_trigger_done/

koriakin: Fixed.

PR #127, opened 2017-04-05
added installer on windows

adamiwaniuk:


altran01:
By the way we might want to clean up all the unnecessary stuff (like other .exe files, and .ilk/.lib files) while working on this installer.

altran01:
Another comment - I don't think user should be able to deselect `Veles UI` component during installation, since it is the only component and it doesn't make sense to continue with installer without it.

PR #129, opened 2017-04-05
Transaction fixes.

koriakin:


python/veles/server/transaction.py
```diff
@@ -105,11 +106,11 @@ def __exit__(self, exc, val, tb):
```
altran01: //I can't write at correct lines since they are out of diff for this PR
I ran pylint on our code and looked through those files and I've got following errors that look true:
E: 97,23: Instance of 'Transaction' has no 'db' member (no-member)
W: 98,20: Using possibly undefined loop variable 'node' (undefined-loop-variable)
E:124, 8: Instance of 'dict' has no 'add' member (no-member)

koriakin: Fixed.

PR #132, opened 2017-04-06
C++ client library - connection management

jacek-izykowski:


src/client/networkclient.cc
```diff
@@ -0,0 +1,371 @@
...
+  message_handlers_["mthd_res"] = &NetworkClient::handleMthdResMessage;
+  message_handlers_["plugin_trigger_run"] = &NetworkClient::handlePluginTriggerRunMessage;
+  message_handlers_["modify_err"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["list_reply"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["get_data_reply"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["ack"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["get_bindata_reply"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["obj_gone"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["create"] = &NetworkClient::handleTreeModificationMessage;
+  message_handlers_["conn_error"] = &NetworkClient::handleConnErrorMessage;
```
altran01: Wrong message type here - should be `connection_error`.
Overall it looks like more of those are wrong as well.

src/client/networkclient.cc
```diff
@@ -0,0 +1,371 @@
...
+  if (client_socket_ && client_socket_->isValid()) {
+    msgpack::sbuffer buf;
+    msgpack::packer<msgpack::sbuffer> packer(buf);
+    messages::MsgpackWrapper::dumpObject(packer, msg);
+    client_socket_->write(buf.data(), buf.size());
+  }
+}
+
+void NetworkClient::registerMessageHandlers() {
+  message_handlers_["sub_cancelled"] = &NetworkClient::handleTreeModificationMessage;
```
altran01: Why does this has the same handler as i.e. `get_reply` - this message doesn't correspond to any tree modification, and is only ack to request to sub cancellation.

jacek-izykowski: Fixed.

src/client/networkclient.cc
jacek-izykowski: Fixed.

src/client/networkclient.cc
```diff
@@ -0,0 +1,410 @@
...
+          << "    code: " << mpe->err->code.c_str()
+          << "  msg: " << mpe->err->msg.c_str() << endl;
+    }
+
+    disconnect();
+  }
+}
+
+void NetworkClient::handleConnectionsMessage(msg_ptr msg) {
+  // TODO - is this something that client should implement in a subclass?
```
altran01: Maybe at least log something in those stub methods, so we know that we've received message that we don't handle yet?

src/client/networkclient.cc
```diff
@@ -0,0 +1,410 @@
...
+  if (client_socket_) {
+    client_socket_->deleteLater();
+    client_socket_ = nullptr;
+  }
+}
+
+void NetworkClient::newDataAvailable() {
+  while (client_socket_) {
+    if (msg_ptr msg =
+        msgpack_wrapper_.loadMessage(client_socket_)) {
```
altran01: Add some try-catch for `SchemaError` to avoid crash if we get bad message from server?

include/client/networkclient.h
```diff
@@ -0,0 +1,133 @@
...
+
+#include "network/msgpackwrapper.h"
+#include "data/nodeid.h"
+
+namespace veles {
+namespace client {
+
+class NodeTree;
+
+typedef std::pair<bool, std::shared_ptr<std::string>> pair_str;
```
adamiwaniuk: where are used this and next type ?

include/client/networkclient.h
```diff
@@ -0,0 +1,133 @@
...
+  QString client_type_;
+  QByteArray authentication_key_;
+
+  messages::MsgpackWrapper msgpack_wrapper_;
+  std::unordered_map<std::string, MessageHandler> message_handlers_;
+
+  QTextStream* output_stream_;
+};
+
+} // veles
```
adamiwaniuk: these comments are in incorrect order

include/client/node.h
```diff
@@ -0,0 +1,48 @@
...
+typedef std::set<std::string> Tags;
+
+/*****************************************************************************/
+/* Node */
+/*****************************************************************************/
+
+class Node {
+};
+
+} // veles
```
adamiwaniuk: incorrect order of comments

include/client/nodetree.h
```diff
@@ -0,0 +1,40 @@
...
+/* NodeTree */
+/*****************************************************************************/
+
+class NodeTree {
+ public:
+  NodeTree(NetworkClient* network_client);
+  void addRemoteNodeTreeRelatedMessage(msg_ptr msg);
+};
+
+} // client
```
adamiwaniuk: incorrect order of comments

include/client/networkclient.h
jacek-izykowski: pair_str is actually being used in NetworkClient::sendMsgConnect. pair_uint64_t was useless and I have removed it.

src/client/networkclient.cc
jacek-izykowski: Done.

src/client/networkclient.cc
jacek-izykowski: Done.

include/client/networkclient.h
jacek-izykowski: Fixed.

include/client/node.h
jacek-izykowski: Fixed.

PR #135, opened 2017-04-06
async_conn: Move getters and method runner to the connection class.

koriakin:
The eventual goal of this PR series is to get rid of the AsyncNode class.


koriakin:
Fixed.

PR #138, opened 2017-04-06
Exorcise AsyncNode from AsyncTracer.

koriakin:


python/veles/async_conn/tracer.py
```diff
@@ -121,61 +121,60 @@ def get_bindata_size(self, id, key):
         return self._get_from_node(
             id, lambda node: self._get_bindata_size(node, key))
 
-    async def _get_data(self, id, key, adata):
+    async def _get_data(self, node, key, adata):
```
altran01: I understand that despite naming change, it will be `NodeID` instance? (since this is what `Check...` models expect)

koriakin: Yes.

PR #140, opened 2017-04-07
Msgpack nits

mkow:


include/network/msgpackobject.h
```diff
@@ -193,79 +193,74 @@ std::shared_ptr<MsgpackObject> toMsgpackObject(const data::NodeID& val);
...
-    value.push_back(toMsgpackObject(elem));
+std::shared_ptr<MsgpackObject> convertMapHelper(const T& val) {
+  std::map<std::string, std::shared_ptr<MsgpackObject>> value;
+  for (const auto& elem : val) {
+    value[elem.first] = toMsgpackObject(elem.second);
   }
   return std::make_shared<MsgpackObject>(value);
 }
+
+}
```
altran01: Comment about end of namespace as usual?

src/network/msgpackobject.cc
```diff
@@ -156,6 +163,8 @@ void MsgpackObject::fromMsgpack(const msgpack::v2::object &obj) {
     value.ext.first = obj_ext.type();
     value.ext.second = std::make_shared<std::vector<uint8_t>>(obj_ext.data(), obj_ext.data()+obj_ext.size);
     break;
+  default:
```
altran01: Doesn't compile on most platforms, since I added helper variable above (which can't be done if there is another label).

include/network/msgpackobject.h
mkow: Right, fixed.

src/network/msgpackobject.cc
mkow: Fixed.

PR #142, opened 2017-04-08
asrv: Move database management logic to a new DbTracker class.

koriakin:


python/veles/async_client/conn.py
```diff
@@ -22,7 +20,6 @@
     QueryHandler,
     BroadcastHandler,
 )
-from veles.async_conn.node import AsyncNode
 from veles.async_conn.subscriber import (
```
altran01: pylint:
>E: 23, 0: No name 'BaseSubscriberNode' in module 'veles.async_conn.subscriber' (no-name-in-module)
E: 23, 0: No name 'BaseSubscriberData' in module 'veles.async_conn.subscriber' (no-name-in-module)
E: 23, 0: No name 'BaseSubscriberBinData' in module 'veles.async_conn.subscriber' (no-name-in-module)
E: 23, 0: No name 'BaseSubscriberList' in module 'veles.async_conn.subscriber' (no-name-in-module)

python/veles/db/tracker.py
```diff
@@ -0,0 +1,349 @@
...
+    BaseSubscriberList,
+)
+from .backend import DbBackend
+from .transaction import Transaction
+from .node import DbNode
+
+
+class DbTracker(object):
+    def __init__(self, db):
+        if not isinstance(db, DbBackend):
```
altran01: wut? You allow for someone to pass something else than DbBackend instance, and then silently fix it?

python/veles/db/tracker.py
```diff
@@ -0,0 +1,349 @@
...
+        # with all their subscriptions).
+        self.all_subs = {}
+        # WTF H4X: this strange construction does two things:
+        #
+        # - the self.nodes dictionary ensures there can be at most one DbNode
+        #   for every node, so there is only one place to invalidate.
+        # - the lru_cache on get_cached_node ensures at least 128 last nodes
+        #   are actually kept alive.
+        self.nodes = weakref.WeakValueDictionary()
+        self.get_cached_node = lru_cache(maxsize=128)(self._get_cached_node)
```
altran01: Hmm, maybe make this 128 a parameter? Since it will probably one of parameters to fine-tuning the server.

python/veles/db/tracker.py
koriakin: Yeah, so you can do AsyncLocalConnection('x.db') instead of constructing the tracker and backend manually.  Or do we want to do that?

python/veles/async_client/conn.py
koriakin: Fixed.

python/veles/db/tracker.py
koriakin: Made it a module-level constant for now, we'll think of tweaks later.

python/veles/db/backend.py
```diff
@@ -523,7 +523,8 @@ def list(self, parent, tags=frozenset(), pos_filter=PosFilter()):
         return {NodeID(bytes(x)) for x, in c.fetchall()}
 
     def begin(self):
-        assert not self.db.in_transaction
+        if six.PY3:
```
altran01: Hmm, now that I think about it - this is server code, we decided that server will only run on Py3 - then why do we need this check?

koriakin: It just happens to be the only part of this file that had problems running on py2, and I figured I would just fix it if it was that easy.  While it probably won't be useful for any practical purpose, it will increase py2 test coverage (once we have tracker tests) of other areas of the code - that's how I found the __ne__ issue, for one.

PR #143, opened 2017-04-10
Add triggers to the protocol.

koriakin:
Protocol changes only - no implementation yet.

### DIFF
--- a/python/veles/async_conn/tracer.py
+++ b/python/veles/async_conn/tracer.py
@@ -121,6 +121,19 @@ class AsyncTracer:
         return self._get_from_node(
             id, lambda node: self._get_bindata_size(node, key))
 
+    def _get_trigger(self, node, key):
+        res = node.triggers.get(key)
+        self.checks.append(check.CheckTrigger(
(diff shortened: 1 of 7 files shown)


python/veles/async_conn/tracer.py
```diff
@@ -121,6 +121,18 @@ def get_bindata_size(self, id, key):
...
+        res = node.triggers.get(key)
+        self.checks.append(check.CheckTrigger(
+            node=node.id,
+            key=key,
+            state=res,
+        ))
+        return res
+
+    def get_trigger(self, id, key):
+        return self._get_from_node(id, lambda node: self._get_trigger(node, key))
```
altran01: pep8 fails :*(

koriakin: Fixed.

altran01:
`TriggerState` isn't generated in C++ (in PR #159 I'm changing code generation to improve it a bit - including no need to explicitly specify EnumModels to generate).

koriakin:
Fixed.

PR #144, opened 2017-04-10
Updated stylesheet for QTabBars

jacek-izykowski:


mkow:
Clicking is now fixed (no tab resize), but when you try to drag a tab the one on the right of it moves a bit, could you fix also this issue?

altran01:
BTW. while we are on topic of moving tabs around - I've noticed that sometimes when moving tabs around the tab I'm moving around, and sometimes even whole tab bar becomes invisible for a moment. However this is probably a separate issue.

jacek-izykowski:
@mkow Fixed.

PR #146, opened 2017-04-10
Wip/altran/nodeid cpp tests

altran01:


include/data/nodeid.h
```diff
@@ -28,19 +28,25 @@ namespace data {
...
+  NodeID(const std::string& data);
+  NodeID(const NodeID& other);
 
   QString toHexString() const;
-  std::vector<uint8_t> asStdVector() const {
-    return std::vector<uint8_t>(value, value + WIDTH_);
-  }
+  static std::shared_ptr<NodeID> fromHexString(QString& val);
+  std::vector<uint8_t> asStdVector() const;
+  // functions for more convinient getting of special values
```
mkow: conv**e**nient

include/data/nodeid.h
```diff
@@ -28,19 +28,25 @@ namespace data {
 class NodeID {
   static const size_t WIDTH_ = 24;
   uint8_t value[WIDTH_];
+  std::random_device random_;
```
mkow: This should be mt19937 initialized from random_device.

random_device is very slow and we don't need security here.

include/data/nodeid.h
```diff
@@ -28,19 +28,25 @@ namespace data {
...
   QString toHexString() const;
-  std::vector<uint8_t> asStdVector() const {
-    return std::vector<uint8_t>(value, value + WIDTH_);
-  }
+  static std::shared_ptr<NodeID> fromHexString(QString& val);
+  std::vector<uint8_t> asStdVector() const;
+  // functions for more convinient getting of special values
+  static std::shared_ptr<NodeID> getRootNodeId();
+  static std::shared_ptr<NodeID> getNilId();
+  bool operator==(const NodeID &other) const;
```
mkow: `NodeID &` -> `NodeID& `

include/data/nodeid.h
```diff
@@ -28,19 +28,25 @@ namespace data {
...
+  bool operator!=(const NodeID &other) const;
```
mkow: ditto

include/data/nodeid.h
```diff
@@ -28,19 +28,25 @@ namespace data {
...
  public:
   static const uint8_t NIL_VALUE[WIDTH_];
-  // TODO singleton
   static const uint8_t ROOT_VALUE[WIDTH_];
 
   NodeID();
   NodeID(const uint8_t* data);
-  NodeID(const QString& data);
+  NodeID(const std::string& data);
+  NodeID(const NodeID& other);
```
mkow: add `explicit` to all 1-arg constructors

src/data/nodeid.cc
```diff
@@ -14,21 +14,61 @@ const uint8_t NodeID::ROOT_VALUE[NodeID::WIDTH_] = {
 };
 
 NodeID::NodeID() {
-  memcpy(value, NIL_VALUE, WIDTH_);
+  for (size_t i = 0; i < WIDTH_; i+=sizeof(unsigned int)) {
+    *reinterpret_cast<unsigned int*>(&value[i]) = random_();
```
mkow: Fix this.
`random_` will be mersenne-twister, so I guess we can fill it byte-by-byte with decent performance.

src/data/nodeid.cc
```diff
@@ -14,21 +14,61 @@ const uint8_t NodeID::ROOT_VALUE[NodeID::WIDTH_] = {
...
 
+NodeID::NodeID(const NodeID &other) {
+  memcpy(value, other.value, WIDTH_);
+}
+
 QString NodeID::toHexString() const {
   return HexEncoder().encode(value, WIDTH_);
 }
 
+std::shared_ptr<NodeID> NodeID::fromHexString(QString &val) {
```
mkow: `NodeID &val` -> `const NodeID& val`

src/data/nodeid.cc
```diff
@@ -14,21 +14,61 @@ const uint8_t NodeID::ROOT_VALUE[NodeID::WIDTH_] = {
...
+  if (val.size() != 48) {
```
mkow: Don't use magic constants. Use width*2.

src/network/msgpackobject.cc
```diff
@@ -380,7 +380,7 @@ std::shared_ptr<MsgpackObject> toMsgpackObject(const double val) {
 
 void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, std::shared_ptr<data::NodeID>& out) {
```
mkow: Why this is void function returning result by arg?

mkow: Same for other fromMsgpackObject functions.

src/network/msgpackobject.cc
```diff
@@ -443,5 +443,16 @@ void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, double& out) {
   out = obj->getDouble();
 }
 
+namespace details_ {
+
+std::shared_ptr<MsgpackObject> details_::convertNodeIDHelper(const data::NodeID &val) {
```
mkow: `const data::NodeID &val` -> `const data::NodeID& val`

test/data/nodeid.cc
```diff
@@ -0,0 +1,64 @@
...
+  EXPECT_THAT(n1, AllOf(Ne(n2), Ne(n3), Ne(n4)));
+  EXPECT_THAT(n2, AllOf(Ne(n1), Ne(n3), Ne(n4)));
+  EXPECT_THAT(n3, AllOf(Ne(n2), Ne(n1), Ne(n4)));
+  EXPECT_THAT(n4, AllOf(Ne(n2), Ne(n3), Ne(n1)));
+
+  NodeID nil(NodeID::NIL_VALUE), root(NodeID::ROOT_VALUE);
+  EXPECT_EQ(nil, *NodeID::getNilId());
+  EXPECT_EQ(root, *NodeID::getRootNodeId());
+
+  NodeID from_str(std::string("000000000000000000000000"));
```
mkow: explicit `std::string()` is not needed here.

test/data/nodeid.cc
```diff
@@ -0,0 +1,64 @@
...
+  };
+  NodeID from_char(data);
+  EXPECT_EQ(from_str, from_char);
+}
+
+TEST(NodeID, ToFromHexString) {
+  QString str(48, QChar('3'));
+  auto id = NodeID::fromHexString(str);
+  EXPECT_EQ(str, id->toHexString());
+  NodeID from_str(std::string("333333333333333333333333"));
```
mkow: ditto

test/data/nodeid.cc
```diff
@@ -0,0 +1,64 @@
...
+  const uint8_t data[] = {
+    0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
+    0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
+  };
+  NodeID from_char(data);
+  EXPECT_EQ(from_str, from_char);
+}
+
+TEST(NodeID, ToFromHexString) {
+  QString str(48, QChar('3'));
```
mkow: Use width*2 instead of 48.

test/data/nodeid.cc
```diff
@@ -0,0 +1,64 @@
...
+  auto id = NodeID::fromHexString(str);
+  EXPECT_EQ(str, id->toHexString());
+  NodeID from_str(std::string("333333333333333333333333"));
+  EXPECT_EQ(*id, from_str);
+  EXPECT_EQ(from_str.toHexString(), str);
+  EXPECT_EQ(NodeID::fromHexString(QString()), nullptr);
+}
+
+TEST(NodeID, ToVector) {
+  EXPECT_THAT(NodeID::getNilId()->asStdVector(), ContainerEq(std::vector<uint8_t>(24, 0)));
```
mkow: Use width instead of 24.

test/data/nodeid.cc
```diff
@@ -0,0 +1,64 @@
...
+  EXPECT_THAT(NodeID::getRootNodeId()->asStdVector(), ContainerEq(std::vector<uint8_t>(24, 255)));
```
mkow: ditto

src/network/msgpackobject.cc
altran01: Will look into it tomorrow.

include/data/nodeid.h
```diff
@@ -26,26 +26,30 @@ namespace veles {
 namespace data {
 
 class NodeID {
+ public:
   static const size_t WIDTH_ = 24;
```
mkow: It's now public, so `WIDTH_` -> `WIDTH`

src/network/msgpackobject.cc
altran01: The only solution for templating return type I could think of looks something like that (due limitations of being unable to do partial function specialization):

```
template <class T>
struct FromMsgpackObject
{
  static T call(const std::shared_ptr<messages::MsgpackObject> obj);
};
template <>
struct FromMsgpackObject<bool>
{
  static bool call(const std::shared_ptr<messages::MsgpackObject> obj) {return true;}
};
template <class S>
struct FromMsgpackObject<std::shared_ptr<std::vector<S>>>
{
  static std::shared_ptr<std::vector<S>> call(const std::shared_ptr<messages::MsgpackObject> obj) {return nullptr;}
};
```
and calling it like this:
```
FromMsgpackObject<bool>::call(nullptr);
FromMsgpackObject<std::shared_ptr<std::vector<bool>>>::call(nullptr);
```
I didn't think this looks better than current version, but I'm open to suggestions.

src/data/nodeid.cc
```diff
@@ -7,67 +7,73 @@ using veles::util::encoders::HexEncoder;
...
+const uint8_t NodeID::ROOT_VALUE[NodeID::WIDTH] = {
   0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
   0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
 };
 
+std::mt19937 NodeID::random_= [](){
+    std::array<int, 5> seed_data;
+    std::random_device r;
+    std::generate_n(seed_data.data(), seed_data.size(), std::ref(r));
+    std::seed_seq seq(std::begin(seed_data), std::end(seed_data));
```
mkow: Why not `seed_data.begin()` and `seed_data.end()`?

altran01: I've copied it from some seeding example - will change.

PR #148, opened 2017-04-10
Log widget is a default active widget

jacek-izykowski:


mkow:
This doesn't work when starting Veles with command line arguments (files to open). In such case the last opened file should be active, but currently all tabs are inactive.

jacek-izykowski:
@mkow Fixed.

src/ui/main.cc
```diff
@@ -20,7 +20,7 @@
 #include <QSurfaceFormat>
 #include <QTranslator>
 #include <QHostAddress>
-
+#include <iostream>
```
altran01: This is already included above

include/ui/veles_mainwindow.h
```diff
@@ -71,7 +71,7 @@ class VelesMainWindow : public MainWindowWithDetachableDockWidgets {
   void createMenus();
   void createDb();
   void createFileBlob(QString);
-  void createHexEditTab(QString, dbif::ObjectHandle);
+  DockWidget* createHexEditTab(QString, dbif::ObjectHandle);
```
altran01: I'm missing how is this supposed to work. I can only find 2 places there createHexEditTab is used, and in both of them nothing is done with return value.

src/ui/main.cc
jacek-izykowski: Fixed.

include/ui/veles_mainwindow.h
jacek-izykowski: Unneeded code removed.

PR #149, opened 2017-04-11
embed python and server srcipts windows installation

adamiwaniuk:


cmake/server.cmake
```diff
@@ -0,0 +1,51 @@
+if(WIN32)
+  set(BASEPYEXE py.exe -3)
+  set(SERVER_DIR ${CMAKE_CURRENT_BINARY_DIR}/veles-server)
+  set(SERVER_PYTHON_DIR ${SERVER_DIR}/python)
+  file(TO_NATIVE_PATH ${SERVER_PYTHON_DIR} SERVER_PYTHON_DIR_NATIVE)
+
+  if(NOT EMBED_PYTHON_ARCHIVE_PATH)
+    set(EMBED_PYTHON_URL "https://www.python.org/ftp/python/3.6.1/python-3.6.1-embed-win32.zip")
```
altran01: Hmm, I've noticed there is also `python-3.6.1-embed-amd64.zip` - since we have 2 Windows distributions (32 and 64bits) can we maybe use 64bit version for 64bit distribution?

adamiwaniuk: fixed

Jenkinsfile
```diff
@@ -132,7 +134,8 @@ builders['msvc2015'] = {node('windows'){
             withEnv(["PATH+QT=${env.QT}\\Tools\\${version}\\bin;${env.QT}\\5.7\\${version}\\bin",
                      "PATH+CMAKE=${env.CMAKE}\\bin",
                      "CMAKE_PREFIX_PATH+QT=${env.QT}\\5.7\\${version}",
-                     "GOOGLETEST_DIR=${tool 'googletest'}"]){
+                     "GOOGLETEST_DIR=${tool 'googletest'}",
+                     "EMBED_PYTHON_ARCHIVE_PATH=${tool 'embed-python-32'}"]){
```
altran01: Those env variables aren't automatically propagated as CMake variables - you can see that `GOOGLETEST_DIR` is used below as cmake execution parameter (I don't even know why it is set as env variable instead of being directly provided there).

adamiwaniuk: fixed

PR #154, opened 2017-04-12
install server script on linux

adamiwaniuk:


resources/install/postinst
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+python3 -m venv /usr/bin/veles-server/venv
```
altran01: Hmm, I don't know if `/usr/bin/` is the best place for this venv, but I don't know where should we put it :(

adamiwaniuk: moved to /usr/share

PR #159, opened 2017-04-14
Muliple changes to code generation:

altran01:
* Split generated code to namespaces named in a manner
  similar to exceptions.h
* Generate code directly to files that will be used, instead of
  defines - this provides more meaningful compile errors, and better
  IDE support
* Instead of explicit list of models to generate in generate.py it now
  takes list of modules to generate from as parameters
* Move definitions of most class methods to source file
* Generate module-level integers as constants in C++ code


python/veles/cpp/generate.py
```diff
@@ -13,81 +13,131 @@
...
-        for sub_class in el.object_types.values():
-            code += sub_class.generate_source_code()
-
-    return code
+    parser = argparse.ArgumentParser()
+    parser.add_argument('destination')
+    parser.add_argument('source_modules', nargs='+')
+    args = parser.parse_args()
+    license = '''/*
+ * Copyright 2017 CodiLime
```
adamiwaniuk: maybe calculate year basing on current data :)

CMakeLists.txt
```diff
@@ -198,11 +198,21 @@ qt5_use_modules(parser Core)
...
-add_custom_command(OUTPUT ${MSGPACK_CPP}
-  COMMAND ${PYEXE} -m veles.cpp.generate > ${MSGPACK_CPP}
+add_custom_command(OUTPUT ${MSGPACK_CPP_HEADER} ${MSGPACK_CPP_SOURCE} ${MSGPACK_CPP_FWD_HEADER}
+  COMMAND ${PYEXE} -m veles.cpp.generate ${CMAKE_CURRENT_BINARY_DIR}
+    veles.proto.node
+    veles.proto.chunk
+    veles.proto.check
+    veles.proto.connection
+    veles.proto.operation
+    veles.proto.messages
```
koriakin: desynced with the list below

PR #163, opened 2017-04-18
create NSIS installer only on non debug builds

adamiwaniuk:


adamiwaniuk:
It is not necessary, only generating windows  installers have huge impact on build time

PR #167, opened 2017-04-20
Wip/mkow/python nits

mkow:
low priority nits.

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -1,8 +1,8 @@
 [metadata]
 name = veles
-home-page = https://codisec.com/veles/
+home-page = https://veles.io/
 author = CodiSec
-author-email = veles@codisec.com
+author-email = contact@veles.io
 
 [files]
 packages = veles
--- a/python/show
+++ b/python/show
@@ -4,10 +4,8 @@ import sys
 
 from veles.schema import nodeid
 from veles.scli import client
+from pprint import pprint
 
 client = client.create_client(sys.argv[1], sys.argv[2])
-
 obj = nodeid.NodeID.from_hex(sys.argv[3])
-
-import pprint
-pprint.pprint(client.get(obj))
+pprint(client.get(obj))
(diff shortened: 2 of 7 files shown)


python/test_dis
```diff
@@ -14,15 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from veles.dis.isa.falcon import FalconIsa
-from veles.data.bindata import BinData
 import sys
+from pprint import pprint
+from veles.data.bindata import BinData
```
altran01: If you are so strict about fixing nits include an empty line between stdlib and local imports :)

python/setup.cfg
```diff
@@ -1,8 +1,8 @@
 [metadata]
 name = veles
-home-page = https://codisec.com/veles/
+home-page = https://veles.io/
 author = CodiSec
-author-email = veles@codisec.com
+author-email = contact@veles.io
```
adamiwaniuk: please update email also in other places

python/test_dis
altran01: Also why is this only file where you moved imports around?

python/sub
```diff
@@ -4,11 +4,9 @@ import sys
...
+from pprint import pprint
 
 client = client.create_client(sys.argv[1], sys.argv[2])
-
 obj = nodeid.NodeID.from_hex(sys.argv[3])
-
-import pprint
 for x in client.get_sub(obj):
-	pprint.pprint(x)
+	pprint(x)
```
altran01: When you are touching lines with tabulation can you change it to spaces as well :) ?

altran01: `python/listsub` has construct identical to what you changed here - any reason you didn't do it there?

python/show
```diff
@@ -4,10 +4,8 @@ import sys
 
 from veles.schema import nodeid
 from veles.scli import client
+from pprint import pprint
```
altran01: Nit. order imports according to PEP8.

python/test_dis
mkow: Will fix in next PR.

PR #173, opened 2017-04-21
UI/UX improvements in 3D view

mkow:


src/visualization/manipulator.cc
```diff
@@ -288,7 +290,6 @@ bool FreeManipulator::mouseEvent(QObject* watched, QMouseEvent* event) {
         }
         prev_pos_ = event->pos();
       }
-      return true;
```
altran01: Why those changes to don't filter out those handled mouse events (we still filter out keyboard/wheel though)?

src/visualization/trigram.cc
```diff
@@ -305,16 +312,15 @@ void TrigramWidget::setManipulator(Manipulator* manipulator) {
 
 void TrigramWidget::autoSetBrightness() {
   auto new_brightness = suggestBrightness();
-  if (new_brightness == brightness_) return;
-  brightness_ = new_brightness;
-  if (brightness_slider_ != nullptr) {
```
altran01: Eh, I'm not a great fan of removing braces (but not a big deal either way).

src/visualization/manipulator.cc
```diff
@@ -100,14 +101,13 @@ QString TrackballManipulator::manipulatorName() {
 
 bool TrackballManipulator::eventFilter(QObject* watched, QEvent* event) {
   QWidget* widget = dynamic_cast<QWidget*>(watched);
-  if (widget == nullptr) {
```
adamiwaniuk: why you changed that?

src/visualization/manipulator.cc
mkow: As discussed privately: it doesn't matter in this case, but we may change this in the future.

src/visualization/trigram.cc
mkow: I prefer to not have them in such shorts and simple ifs (IMO: no gain in readability if indentation is correct, but one line longer). I probably won't touch them in future PRs ;)

src/visualization/manipulator.cc
mkow: I thought that Google style guide prefers `!widget`, but seems that's true only for Python style guide. Keep it or revert?

mkow:
@altran01: seems that initial lag was there before this commit, but was harder to notice because of the glitches.

PR #178, opened 2017-04-24
Allow for transfer of Repacker class over network

altran01:
This includes changes needed to create model that will have class
with additional custom code in C++ that inherits from generated code

### DIFF
--- a/include/data/repack.h
+++ b/include/data/repack.h
@@ -17,27 +17,17 @@
 #pragma once
 
 #include "data/bindata.h"
+#include "models.h"
 
 namespace veles {
 namespace data {
 
-/** Represents endianness used for repacking data.  */
-enum class Endian {
-  LITTLE, BIG,
-};
-
 /** Represents data repacking format.  */
-struct Repacker {
-  /** Endanness to use for repacking.  */
-  Endian endian;
-  /** Element width, in bits, to repack from.  */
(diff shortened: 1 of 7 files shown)


include/data/repack.h
```diff
@@ -17,27 +17,17 @@
...
-  unsigned from_width;
-  /** Element width, in bits, to repack into.  */
-  unsigned to_width;
-  /** Amount of padding, in bits, to be skipped around each element.
-      Low pad is between LSB and the actual element, high pad
-      is between MSB and the actual element.  */
-  unsigned high_pad, low_pad;
+class Repacker : public RepackerModel {
+ public:
+  Repacker(Endian endian = Endian::LITTLE, uint64_t from_width = 1, uint64_t to_width = 1, uint64_t high_pad = 0, uint64_t low_pad = 0) :
```
adamiwaniuk: why these default values ? in current code for exmaple for from_width 8 is used most of the time

altran01: {high,low}_pad have defaults from Python code, for {from,to}_width I chose minimal value that makes any sense (0 doesn't) - I can change them to 8 if you prefer.

adamiwaniuk: Yes, I prefer 8, because right now it is used most frequently in code

PR #184, opened 2017-04-26
Add loading/packing of BinData in C++ code

altran01:


src/network/msgpackobject.cc
```diff
@@ -428,6 +428,15 @@ std::shared_ptr<MsgpackObject> toMsgpackObject(const std::shared_ptr<data::NodeI
   return details_::convertNodeIDHelper(*val);
 }
 
+std::shared_ptr<MsgpackObject> toMsgpackObject(const std::shared_ptr<data::BinData> val) {
+  auto data = std::make_shared<std::vector<uint8_t>>();
+  for (int i = 0; i < 4; ++i) {
+    data->push_back((val->width() << i * 8) & 0xff);
```
adamiwaniuk: please add () to clarify priority of << and *

koriakin: should be >>

src/network/msgpackobject.cc
```diff
@@ -452,6 +461,16 @@ void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, std::shared_ptr
   }
 }
 
+void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, std::shared_ptr<data::BinData>& out) {
+  if (obj->getExt().first != proto::EXT_BINDATA) {
+    throw proto::SchemaError("Wrong ext type for BinData");
+  }
+  auto data = obj->getExt().second;
+  uint32_t width = (data->at(3) << 24) | (data->at(2) << 16) | (data->at(1) << 8) | (data->at(0));
```
koriakin: that (and the above) sounds like a job  for veles::util

src/network/msgpackobject.cc
```diff
@@ -452,6 +461,16 @@ void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, std::shared_ptr
   }
 }
 
+void fromMsgpackObject(const std::shared_ptr<MsgpackObject> obj, std::shared_ptr<data::BinData>& out) {
+  if (obj->getExt().first != proto::EXT_BINDATA) {
+    throw proto::SchemaError("Wrong ext type for BinData");
+  }
+  auto data = obj->getExt().second;
```
koriakin: no validation of data size - bad things happen if size is < 4

include/util/misc.h
```diff
@@ -0,0 +1,43 @@
...
+#pragma once
+
+namespace veles {
+namespace util {
+
+template<typename T>
+T bytesToIntLe(const uint8_t* bytes, size_t num_bytes) {
+  assert(num_bytes <= sizeof(T));
+  T res = T();
+  for (int i = 0; i < num_bytes; ++i) {
```
adamiwaniuk: warning?

include/util/misc.h
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ *
+ */
+#pragma once
+
+namespace veles {
+namespace util {
+
+template<typename T>
+T bytesToIntLe(const uint8_t* bytes, size_t num_bytes) {
```
adamiwaniuk: in my opinion it will be better to use here as input vector<uint8_t> and the same as output on second function

include/util/misc.h
```diff
@@ -0,0 +1,43 @@
...
+  assert(num_bytes <= sizeof(T));
+  T res = T();
+  for (int i = 0; i < num_bytes; ++i) {
+    res |= static_cast<T>(bytes[i]) << (i * 8);
+  }
+  return res;
+}
+
+template<typename T>
+uint8_t* intToBytesLe(T val, size_t num_bytes) {
```
koriakin: Do we want this to alloc memory? IMO would be better to store bytes to caller-allocated storage.

altran01: Do you prefer that we use std::vector here (as @adamiwaniuk suggested - I was thinking about it, but I wasn't sure whether we will always use std::vector or i.e. QVector), or stick to simple array, but allocated by caller?

koriakin: Just use pointers in both cases.

CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ add_library(veles_base
 
     ${INCLUDE_DIR}/util/string_utils.h
     ${INCLUDE_DIR}/util/math.h
+    ${INCLUDE_DIR}/util/misc.h
```
koriakin: "util/misc" is a useless name, call it util/int_bytes

PR #186, opened 2017-04-27
build mingw only on release builds

adamiwaniuk:


adamiwaniuk:
Build time: "Took 6 min 17 sek"

PR #190, opened 2017-05-04
added network profile support

adamiwaniuk:
this PR contains only settings change, next one will include UI

### DIFF
--- a/src/util/settings/connection_client.cc
+++ b/src/util/settings/connection_client.cc
@@ -30,19 +30,67 @@ QString localhost("127.0.0.1");
 QString default_database_file("veles.vdb");
 int default_server_port = 3135;
 bool default_shut_down_server = true;
+QString default_profile("local server");
+
+QString currentProfile() {
+  QSettings settings;
+  return settings.value("profile", default_profile).toString();;
+}
+
+void setCurrentProfile(QString profile) {
+  QSettings settings;
+  settings.setValue("profile", profile);
+}
+
+void removeProfile(QString profile) {
+    QSettings settings;
+    QMap<QString, QVariant> profiles = settings.value("profiles").toMap();
+    profiles.remove(profile);
+    settings.setValue("profiles", profiles);
+    if (profile == currentProfile()) {
+      auto profiles = profileList();
+      if (profiles.empty()) {
+        setCurrentProfile(default_profile);
+      } else {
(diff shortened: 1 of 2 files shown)


src/util/settings/connection_client.cc
```diff
@@ -30,67 +30,108 @@ QString localhost("127.0.0.1");
...
+  return settings.value("profile", default_profile).toString();;
+}
+
+void setCurrentProfile(QString profile) {
+  QSettings settings;
+  settings.setValue("profile", profile);
+}
+
+void removeProfile(QString profile) {
+    QSettings settings;
```
adamiwaniuk: niespójne wcięcia

src/util/settings/connection_client.cc
```diff
@@ -30,75 +30,116 @@ QString localhost("127.0.0.1");
...
+  QSettings settings;
+  return settings.value("profile", default_profile).toString();;
+}
+
+void setCurrentProfile(QString profile) {
+  QSettings settings;
+  settings.setValue("profile", profile);
+}
+
+void removeProfile(QString profile) {
```
altran01: Hmm, there is option to remove a profile, but I don't see any to add one?

adamiwaniuk: when setCurrentProfile is called with non exisiting profile, then it will be created while setting any setting

altran01: Mhm, I thought that invalid QVariant (default constructed) can't be converted to map, but it seems that I was wrong.

src/util/settings/connection_client.cc
```diff
@@ -30,75 +30,116 @@ QString localhost("127.0.0.1");
...
+  QMap<QString, QVariant> default_profiles;
+  default_profiles[default_profile] = QSettings::SettingsMap();
+  QMap<QString, QVariant> profiles = settings.value("profiles", default_profiles).toMap();
+  return profiles.keys();
+}
+
+QVariant profileSettings(QString key, QVariant defaultValue) {
+  QSettings settings;
+  QMap<QString, QVariant> profiles = settings.value("profiles", QSettings::SettingsMap()).toMap();
+  QSettings::SettingsMap proefileSettings = profiles.value(currentProfile(), QSettings::SettingsMap()).toMap();
```
adamiwaniuk: fixed

PR #191, opened 2017-05-04
added profile support in ui

adamiwaniuk:


src/ui/connectiondialog.cc
```diff
@@ -231,18 +240,26 @@ void ConnectionDialog::saveSettings() {
       ui_->server_executable_line_edit->text());
   util::settings::connection::setShutDownServerOnQuit(
       ui_->shut_down_server_check_box->isChecked());
-}
 
-void ConnectionDialog::saveSettingsToggled(bool toggled) {
-  ui_->save_key_check_box->setEnabled(toggled);
```
altran01: You removed this function which was called when removed `save_settings_check_box` was used, which causes that `save_key_check_box` is always disabled and unusable.

src/ui/connectionmanager.cc
```diff
@@ -318,21 +319,27 @@ void ConnectionNotificationWidget::updateConnectionStatus(
     last_status_change_ = frame_;
   }
 
+  QString status_text;
   switch (connection_status) {
   case client::NetworkClient::ConnectionStatus::NotConnected:
-    ui_->connection_status_text_label->setText("Not connected");
+    status_text = "Not connected";
```
altran01: Hmm, I don't know if displaying `Not connected:<currently selected profile>` is what I'd expect.

src/ui/connectiondialog.cc
```diff
@@ -188,6 +189,14 @@ void ConnectionDialog::loadDefaultValues() {
       util::settings::connection::shutDownServerOnQuitDefault());
 }
 
+void ConnectionDialog::loadProfiles() {
```
altran01: nit. if we have deleted our last profile `local server` will be entered in QComboBox, but it won't be selectable element of this QComboBox.

adamiwaniuk: fixed

src/ui/connectiondialog.cc
adamiwaniuk: fixed

adamiwaniuk:
also improved UX

altran01:
Some more UX comments from my point of view (we can discuss it if you don't agree):

- I'd like to either be able to connect to server without saving it OR add some checkbox to prevent auto connecting to in on next run.
- user can silently overwrite one of existing profiles by selecting already taken name (this can lead to lost auth-keys)

adamiwaniuk:
added some more UX improvments

PR #195, opened 2017-05-08
wip/ji/dbif-parsers

jacek-izykowski:


altran01:
With parsers logging so much information in `NCWrapper` for each request absolutely kills it's performance, because it tries to write it everything to log widget. (with proper logging we could set it to some TRACE log level, but for now I think our best option is to simply remove it).

jacek-izykowski:
@altran01 I agree. detailed_debug_info_ is now set to false by default. In the future we might add a possibility to turn it on/off from options dialog.

altran01:
Eh, even with above fix, performance drop for pyc parser is abysmal at best - .pyc file that before took < 10seconds to parse now it took around 40minutes. From brief analysis it looks like the main culprit is huge amount of small transactions in sqlite which can't handle it in timely manner (CPU usage for both Python and C++ processes was negligible, while Python constantly used disk I/O on sqlite db and journal files).

src/client/dbif.cc
```diff
@@ -565,25 +646,41 @@ dbif::InfoPromise* NCWrapper::handleBlobDataRequest(
 }
 
 dbif::InfoPromise* NCWrapper::handleChunkDataRequest(data::NodeID id, bool sub) {
-  uint64_t qid = nc_->nextQid();
+  uint64_t qid_data = nc_->nextQid();
+  uint64_t qid_children = nc_->nextQid();
```
altran01: Hmm, can we make `nextQid` use some atomic counter inside to avoid risk of duplicate qid if we have situation where two threads use it concurrently?

PR #200, opened 2017-05-09
Wip/mkow/connection dialog fix + nits

mkow:


src/util/settings/connection_client.cc
```diff
@@ -213,14 +214,16 @@ QString serverScriptDefault() {
   server_script = qApp->applicationDirPath() + "/../veles-server/srv.py";
 #elif defined(Q_OS_LINUX)
   server_script = qApp->applicationDirPath() + "/../share/veles-server/srv.py";
-#else
+#elif defined(Q_OS_MAC)
   server_script = qApp->applicationDirPath() + "/../Resources/veles-server/srv.py";
+#else
+  #error OS not supported
```
altran01: If you are going to add such error then maybe we should also do it in other similar places - i.e. `src/ui/connectionmanager.cc` ?

mkow: IMO it's ok in other places. E.g.: in the file you linked to, the if checks whether we are working on Windows and then does something special, otherwise does something else. No reason to list specific OSes there, as it's only there to support Windows as a special case.

altran01: Sorry I wasn't specific enough - I meant situation in this ifdef sequence https://github.com/codilime/veles/blob/master/src/ui/connectionmanager.cc#L175.

mkow: My bad, I grepped for "#ifdef" ;) Fixed.

adamiwaniuk: I don't like this change, if someone is using other os it should still work on version builded from sources, but maybe it is worth to change order and put on #else linux version

src/ui/connectionmanager.cc
```diff
@@ -184,12 +184,14 @@ void ConnectionManager::startLocalServer() {
...
   }
-#else
+#elif defined(Q_OS_MAC)
   QString python_interpreter_executable(qApp->applicationDirPath() + "/../Resources/veles-server/venv/bin/python3");
   QFileInfo check_file(python_interpreter_executable);
   if (!check_file.exists()) {
     python_interpreter_executable = QString("python3");
   }
+#else
+  #error OS not supported
```
adamiwaniuk: why? without this change it is chance that it will still work on other OS because of fallback to "python3"

src/util/settings/connection_client.cc
mkow: But which path should we use for that 'other OS'? Before this change it would just pick the random one which happened to be in `#else`, which almost for sure won't work. This way someone compiling this for exotic OSes will explicitly have to fix the paths here.

src/ui/connectionmanager.cc
mkow: IMO it's still better to fail explicitly here, so user will have to fix it for strange OSes. Also, in case we port Veles to something other we won't forget to fix this place.

src/util/settings/connection_client.cc
mkow: check now

src/ui/connectionmanager.cc
mkow: check now

PR #203, opened 2017-05-10
wip/ji/command-line-files-opening

jacek-izykowski:


include/ui/veles_mainwindow.h
```diff
@@ -104,6 +104,9 @@ class VelesMainWindow : public MainWindowWithDetachableDockWidgets {
   ConnectionNotificationWidget* connection_notification_widget_;
 
   QToolBar* tool_bar_;
+
+  bool have_ever_been_connected_;
```
altran01: super nit. has?

src/ui/veles_mainwindow.cc
```diff
@@ -281,6 +284,23 @@ void VelesMainWindow::updateConnectionStatus(
       == client::NetworkClient::ConnectionStatus::Connected) {
     database_dock_widget_->widget()->setEnabled(true);
     open_act_->setEnabled(true);
+
+    if (!have_ever_been_connected_) {
```
adamiwaniuk: do we realy need this flag here, why do not use files_to_upload_once_connected_.empty() ?

src/ui/veles_mainwindow.cc
```diff
@@ -45,13 +45,16 @@ namespace ui {
...
-    database_dock_widget_(0), log_dock_widget_(0) {
+    database_dock_widget_(0), log_dock_widget_(0),
+    have_ever_been_connected_(false) {
   setAcceptDrops(true);
   resize(1024, 768);
   init();
 }
 
-void VelesMainWindow::addFile(QString path) { createFileBlob(path); }
+void VelesMainWindow::addFile(QString path) {
```
adamiwaniuk: now drag and drop on already connected veles is broken

include/client/networkclient.h
```diff
@@ -135,3 +135,5 @@ public slots:
 
 }  // namespace client
 }  // namespace veles
+
+Q_DECLARE_METATYPE(veles::client::NetworkClient::ConnectionStatus)
```
altran01: To do it  "by the books" you should probably also add `qRegisterMetaType` call for this type (it seems to work now without it, but I'd like to avoid situation where on one specific version of Qt it doesn't work) - 

> Note that if you intend to use the type in queued signal and slot connections or in QObject's property system, you also have to call qRegisterMetaType() since the names are resolved at runtime.

adamiwaniuk: why we need this ?

include/ui/veles_mainwindow.h
jacek-izykowski: I assume 1st person here. And I'm going to remove this flag.

include/client/networkclient.h
jacek-izykowski: To use queued signal-slot connections with veles::client::NetworkClient::ConnectionStatus as a parameter.

src/ui/veles_mainwindow.cc

src/ui/veles_mainwindow.cc
jacek-izykowski: Good point. Improved.

include/client/networkclient.h
jacek-izykowski: It works with Q_DECLARE_METATYPE on Qt 5.5. And it doesn't work without it.

src/ui/veles_mainwindow.cc
```diff
@@ -51,7 +51,9 @@ VelesMainWindow::VelesMainWindow() : MainWindowWithDetachableDockWidgets(),
...
```
mkow:
```C++
const QString& path
```

src/ui/veles_mainwindow.cc
```diff
@@ -281,6 +283,19 @@ void VelesMainWindow::updateConnectionStatus(
       == client::NetworkClient::ConnectionStatus::Connected) {
     database_dock_widget_->widget()->setEnabled(true);
     open_act_->setEnabled(true);
+
+    if (!files_to_upload_once_connected_.empty()) {
+      QTextStream out(LogWidget::output());
+      out << "Uploading files specified as command line arguments:" << endl;
+
+      for (auto path : files_to_upload_once_connected_) {
```
mkow:
```C++
const auto& path
```

include/client/networkclient.h
jacek-izykowski: Fixed.

src/ui/veles_mainwindow.cc
jacek-izykowski: Changed.

src/ui/veles_mainwindow.cc
jacek-izykowski: Changed.

PR #212, opened 2017-05-12
Hash function for NodeID

jacek-izykowski:


include/data/nodeid.h
```diff
@@ -54,5 +54,9 @@ class NodeID {
   explicit operator bool() const;
 };
 
+struct NodeIDHash {
```
altran01: Hmm, any reason why not provide specialization for std::hash instead? If we did that I think we wouldn't need to explicitly provide it as parameter to std unordered containers.

jacek-izykowski: This is a matter of taste. Specialization of std::hash requires extending std namespace.

mkow: https://google.github.io/styleguide/cppguide.html#std_hash
```
Do not define specializations of std::hash.
```
Seems this is discouraged.

PR #215, opened 2017-05-15
Add RC notifications instead of slack ones

altran01:
To do it we use hubot-jenkins-notifier plugin for hubot - our RC bot

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,12 +33,12 @@ def getWorkspace = { buildType ->
 }
 
 @NonCPS
-def post_stage_failure(job, stage, error, url){
-  slackSend channel:"eax-builds", color: "danger", message: "Build ${job} failed in stage ${stage} with ${error}, see: <${url}|results>"
+def post_stage_failure(job, stage, build_id, url){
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -33,12 +33,12 @@ def getWorkspace = { buildType ->
 }
 
 @NonCPS
-def post_stage_failure(job, stage, error, url){
-  slackSend channel:"eax-builds", color: "danger", message: "Build ${job} failed in stage ${stage} with ${error}, see: <${url}|results>"
+def post_stage_failure(job, stage, build_id, url){
+  httpRequest url: "http://10.5.88.46:32768/hubot/jenkins-notify?room=eax-builds", httpMode: "POST", contentType: "APPLICATION_JSON", requestBody: "{\"build\": {\"phase\": \"FINISHED\", \"status\": \"FAILURE\", \"number\": ${build_id}, \"full_url\": \"${url}\"}, \"name\": \"${job} stage: ${stage}\" }"
```
adamiwaniuk: could you hide this url in some jenkins configuration variable or something ?

altran01: Eh unfortunately there doesn't seem to be any straight forward way to do it (Global properties plugin seems to be ignored if we use Jenkinsfile).

PR #216, opened 2017-05-16
Add keyboard shortcuts editor

altran01:
Keyboard shortcuts for all action types are saved in settings,
they are editable through new dialog available in file menu.
All existing QActions that weren't generated based on some data
were converted to support editing shortcuts for them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ add_library(veles_base
     ${INCLUDE_DIR}/util/sampling/fake_sampler.h
     ${INCLUDE_DIR}/util/settings/connection_client.h
     ${INCLUDE_DIR}/util/settings/hexedit.h
+    ${INCLUDE_DIR}/util/settings/shortcuts.h
     ${INCLUDE_DIR}/util/settings/theme.h
     ${INCLUDE_DIR}/util/settings/visualization.h
 
@@ -98,6 +99,7 @@ add_library(veles_base
     ${SRC_DIR}/util/sampling/fake_sampler.cc
     ${SRC_DIR}/util/settings/connection_client.cc
     ${SRC_DIR}/util/settings/hexedit.cc
+    ${SRC_DIR}/util/settings/shortcuts.cc
     ${SRC_DIR}/util/settings/theme.cc
     ${SRC_DIR}/util/settings/visualization.cc
 
@@ -308,15 +310,17 @@ target_link_libraries(unpyc veles_db parser veles_network)
 # Resources
 qt5_add_resources(RESOURCES resources/veles.qrc)
 qt5_wrap_ui(FORMS
-    ${SRC_DIR}/ui/optionsdialog.ui
-    ${SRC_DIR}/ui/searchdialog.ui
-    ${SRC_DIR}/ui/gotoaddressdialog.ui
-    ${SRC_DIR}/ui/createchunkdialog.ui
-    ${SRC_DIR}/ui/databaseinfo.ui
-    ${SRC_DIR}/ui/logwidget.ui
-    ${SRC_DIR}/ui/connectiondialog.ui
-    ${SRC_DIR}/ui/connectionnotificationwidget.ui
-    )
+  ${SRC_DIR}/ui/connectiondialog.ui
+  ${SRC_DIR}/ui/connectionnotificationwidget.ui
+  ${SRC_DIR}/ui/createchunkdialog.ui
+  ${SRC_DIR}/ui/databaseinfo.ui
+  ${SRC_DIR}/ui/gotoaddressdialog.ui
+  ${SRC_DIR}/ui/logwidget.ui
+  ${SRC_DIR}/ui/optionsdialog.ui
+  ${SRC_DIR}/ui/searchdialog.ui
+  ${SRC_DIR}/ui/shortcutselection.ui
+  ${SRC_DIR}/ui/shortcutssettings.ui
+)
 
 if(WIN32)
   set(ICONS ${CMAKE_SOURCE_DIR}/resources/icons/veles.rc)
@@ -360,6 +364,8 @@ add_executable(main_ui
     ${INCLUDE_DIR}/ui/spinbox.h
     ${INCLUDE_DIR}/ui/spinboxvalidator.h
     ${INCLUDE_DIR}/ui/connectiondialog.h
+    ${INCLUDE_DIR}/ui/shortcutedit.h
+    ${INCLUDE_DIR}/ui/shortcutssettings.h
     ${SRC_DIR}/ui/main.cc
     ${SRC_DIR}/ui/logwidget.cc
     ${SRC_DIR}/ui/veles_mainwindow.cc
@@ -382,6 +388,8 @@ add_executable(main_ui
     ${SRC_DIR}/ui/spinbox.cc
     ${SRC_DIR}/ui/spinboxvalidator.cc
     ${SRC_DIR}/ui/connectiondialog.cc
+    ${SRC_DIR}/ui/shortcutedit.cc
+    ${SRC_DIR}/ui/shortcutssettings.cc
     ${RESOURCES}
     ${VISUALIZATION_SHADERS}
     ${FORMS}
--- a/include/util/settings/shortcuts.h
+++ b/include/util/settings/shortcuts.h
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2017 CodiLime
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#pragma once
+
+#include <QAbstractItemModel>
+#include <QAction>
+#include <QDataStream>
+#include <QDialog>
+#include <QKeySequence>
+#include <QMenu>
+#include <QMetaType>
+#include <QPointer>
+#include <QString>
+
+namespace veles {
+namespace util {
+namespace settings {
+namespace shortcuts {
+
+enum ShortcutType : uint32_t {
+  UNKNOWN_SHORTCUT = 0,
+  EXIT_APPLICATION = 1,
+  OPEN_FILE = 2,
+  NEW_FILE = 3,
+  SHOW_DATABASE = 4,
+  SHOW_LOG = 5,
+  SHOW_ABOUT = 6,
+  SHOW_ABOUT_QT = 7,
+  SHOW_OPTIONS = 8,
+  SHOW_SHORTCUT_OPTIONS = 9,
+  SHOW_CONNECT_DIALOG = 10,
+  DISCONNECT_FROM_SERVER = 11,
+  KILL_LOCAL_SERVER = 12,
+  DOCK_MOVE_TO_TOP = 13,
+  DOCK_MOVE_TO_TOP_MAX = 14,
+  DOCK_SPLIT_HORIZ = 15,
+  DOCK_SPLIT_VERT = 16,
+  CREATE_CHUNK = 17,
+  CREATE_CHILD_CHUNK = 18,
+  GO_TO_ADDRESS = 19,
+  REMOVE_CHUNK = 20,
+  SAVE_SELECTION_TO_FILE = 21,
+  HEX_FIND = 22,
+  HEX_FIND_NEXT = 23,
+  OPEN_VISUALIZATION = 24,
+  OPEN_HEX = 25,
+  SHOW_NODE_TREE = 26,
+  SHOW_MINIMAP = 27,
+  VISUALIZATION_DIGRAM = 28,
+  VISUALIZATION_TRIGRAM = 29,
+  VISUALIZATION_LAYERED_DIGRAM = 30,
+  VISUALIZATION_OPTIONS = 31,
+  TRIGRAM_CUBE = 32,
+  TRIGRAM_CYLINDER = 33,
+  TRIGRAM_SPHERE = 34,
+  VISUALIZATION_MANIPULATOR_SPIN = 35,
+  VISUALIZATION_MANIPULATOR_TRACKBALL = 36,
+  VISUALIZATION_MANIPULATOR_FREE = 37,
+  COPY = 38,
+};
+
+QMap<ShortcutType, QList<QKeySequence>> defaultShortcuts();
+QList<QKeySequence> getShortcuts(ShortcutType type);
+void setShortcuts(ShortcutType type, const QList<QKeySequence>& shortcuts);
+
+class ShortcutsItem {
+ public:
+  explicit ShortcutsItem(const QString& display_name = QString(""), ShortcutsItem* parent = 0);
+  ShortcutsItem(const QString& display_name, ShortcutsItem* parent, util::settings::shortcuts::ShortcutType type);
+  ShortcutsItem(const QString& name, const QString& display_name, ShortcutsItem* parent, util::settings::shortcuts::ShortcutType type);
+  ~ShortcutsItem();
+
+  QAction* createQAction(QObject* parent, Qt::ShortcutContext context = Qt::WindowShortcut);
+  QAction* createQAction(QObject* parent, QIcon icon, Qt::ShortcutContext context = Qt::WindowShortcut);
+  QString name() const;
+  QString displayName() const;
+  ShortcutsItem* parent() const;
+  QList<QKeySequence> shortcuts() const;
+  QString displayShortcuts() const;
+  bool deleteShortcut(const QKeySequence& shortcut);
+  bool addShortcut(const QKeySequence& shortcut);
+  void setConflict(bool conflict);
+  bool hasConflict() const;
+  util::settings::shortcuts::ShortcutType type() const;
+  void updateShortcutsForActions();
+  bool isCategory() const;
+
+  QList<ShortcutsItem*> children;
+
+ private:
+  QList<QPointer<QAction>> actions_;
+  QList<QKeySequence> shortcuts_;
+  QString display_shortcuts_;
+  QString name_;
+  QString display_name_;
+  ShortcutsItem* parent_;
+  bool conflict_;
+  util::settings::shortcuts::ShortcutType type_;
+  bool is_category_;
+};
+
+class ShortcutsModel : public QAbstractItemModel {
+ public:
+  static const int COLUMN_INDEX_NAME = 0;
+  static const int COLUMN_INDEX_SHORTCUTS = 1;
+  static const int CATEGORY_ROLE = Qt::UserRole;
+  static const int TYPE_ROLE = Qt::UserRole + 1;
+  static const int SHORTCUTS_ROLE = Qt::UserRole + 2;
+
+  ~ShortcutsModel();
+
+  virtual QModelIndex index(int row, int column, const QModelIndex& parent = QModelIndex()) const override;
+  virtual QModelIndex parent(const QModelIndex& index) const override;
+  virtual int rowCount(const QModelIndex& parent = QModelIndex()) const override;
+  virtual int columnCount(const QModelIndex& parent = QModelIndex()) const override;
+  virtual QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+  QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const override;
+  static ShortcutsModel* getShortcutsModel();
+  QAction* createQAction(util::settings::shortcuts::ShortcutType type, QObject* parent, Qt::ShortcutContext context = Qt::WindowShortcut);
+  QAction* createQAction(util::settings::shortcuts::ShortcutType type, QObject* parent, QIcon icon, Qt::ShortcutContext context = Qt::WindowShortcut);
+  void addShortcut(util::settings::shortcuts::ShortcutType type, const QKeySequence& shortcut);
+  void removeShortcut(util::settings::shortcuts::ShortcutType type, const QKeySequence& shortcut);
+  ShortcutsItem* itemFromIndex(const QModelIndex& index) const;
+  QModelIndex indexFromItem(ShortcutsItem* item) const;
+  QList<ShortcutsItem*> getItemsForSequence(const QKeySequence& sequence);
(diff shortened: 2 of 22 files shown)


wiktorn:
screenshot from 2017-05-16 14-11-00

Few things visible  on this screenshot:
* when searching for commands, we can remove groups that do not have matches
* default size of "Add shortcut" dialog could be a bit wider so full message is shown by default ("Press key without m...")
* why grouping items are in grey? They look like inactive

And few not visible on this screenshot:
* It should be "Reset shortcut to default"  not "shortcuts" (it gives intuition, that it might reset all shortcuts)
* when displaying possible conflicts text should be wrapped (maybe each conflict in separate line)

altran01:
About 1st and 3rd comment - QTreeView filtering doesn't support that use-case easily (I added TODO in code and I'll work on it when I fix other issues) - I changed color (only choosing correct one on dark theme) to distinguish empty categories from items. I think we would still like to make categories more noticeable - but I'll use icon for it (for now temporary one from open).
2nd,4th and 5th fixed (for now wrapping - I think it looks better, but we can discuss).

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,467 @@
...
+  ShortcutsItem* parentItem = child->parent();
+
+  if (parentItem == root_) {
+    return QModelIndex();
+  }
+  return createIndex(parentItem->children.indexOf(child), 0, parentItem);
+}
+
+int ShortcutsModel::rowCount(const QModelIndex& parent) const {
+  if (parent.column() > 0)
```
adamiwaniuk: {} or the same line

mkow: It's ok IMO. I don't like the one-line option, but {} are also fine.

adamiwaniuk:
When pressiong "Windows" (meta) key 
windows

mkow:
Another issues:
* We shouldn't allow meta/win shortcuts on Windows, this key is reserved for the OS.
* Numpad keys don't work. On windows they are treated as different keys than their non-numpad counterparts, but we may want to merge them together (because not everyone has numpad).

altran01:
We can discuss the 1st point when you will be in office (or on chat) - we already had a short discussion about it, and I confirmed that you can create shortcuts with win/meta key (if there doesn't exist system shortcut for it - but the same problem exists i.e. for Alt+F4).
I'll look into 2nd point.

mkow:
* It's possible to edit shortcut without any selection (enter "aaaaaaaaaa" as a filter and hit [Enter])
* It crashes after adding such shortcut: search for "aaaaaaa", hit [Enter], [A], click "Ok", hit [Esc]
* It's possible to bind numlock, capslock, scroll lock and menu. We should discuss if it's a good idea to support this.
* You can "cheat" the key textbox when editing a shortcut using RMB -> Paste (and paste anything into this field)
* There should be a way to Confirm/Cancel the dialog using keyboard. What about [Ctrl]+[Enter] and [Shift]+[Esc]?

altran01:
Ad 1&2 - forgot to test it after I added option to edit by keyboard shortcut - will fix.
Ad 3 - I don't have strong preference - when I looked at what other programs do in shortcut editors, some allow it and some don't. (BTW - I wasn't able bind scroll lock :P)
Ad 4 - forgot about it (it doesn't break anything since I don't actually read what is written there) - will fix.
Ad 5 - which dialog are you talking about ? In the second dialog for choosing shortcut you can use `Tab` key to change focus, and then you can use buttons and/or Esc&enter/return (I thought it was fine enough). However in the first dialog I indeed by mistake removed ability to save changes using keyboard (you can still cancel using escape) - will fix.

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+}
+
+ShortcutsModel* ShortcutsModel::getShortcutsModel() {
+  static ShortcutsModel* ptr = new ShortcutsModel();
+  return ptr;
+}
+
+QAction* ShortcutsModel::createQAction(ShortcutType type, QObject* parent, Qt::ShortcutContext context) {
```
mkow: wrap

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  return type_to_shortcut_[type]->createQAction(parent, context);
+}
+
+QAction* ShortcutsModel::createQAction(ShortcutType type, QObject* parent, QIcon icon, Qt::ShortcutContext context) {
```
mkow: wrap

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+void ShortcutsModel::addShortcut(ShortcutType type, QKeySequence& shortcut) {
+  if (shortcut.isEmpty()) {
+    return;
+  }
+  auto item = type_to_shortcut_.value(type);
+  if (!item) {
+    return;
+  }
+  if (item->addShortcut(shortcut)) {
+    if (sequence_to_shortcut_.contains(shortcut) && !sequence_to_shortcut_[shortcut].contains(item)) {
```
mkow: wrap

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+      if (sequence_to_shortcut_[shortcut].size() > 0) {
+        if (sequence_to_shortcut_[shortcut].size() == 1) {
+          sequence_to_shortcut_[shortcut][0]->setConflict(true);
+        }
+        item->setConflict(true);
+      }
+      sequence_to_shortcut_[shortcut].append(item);
+    } else {
+      sequence_to_shortcut_[shortcut].append(item);
```
mkow: this line is copied in both branches of the if, maybe it'd be better to extract it outside?

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  if (shortcut.isEmpty()) {
+    return;
+  }
+  auto item = type_to_shortcut_.value(type);
+  if (!item) {
+    return;
+  }
+  if (item->deleteShortcut(shortcut)) {
+    sequence_to_shortcut_[shortcut].removeAll(item);
+    if (sequence_to_shortcut_[shortcut].size() == 1 && !checkIfConflicts(sequence_to_shortcut_[shortcut][0])) {
```
mkow: wrap

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+
+ShortcutsModel::ShortcutsModel() : root_(new ShortcutsItem()) {
+  auto global = addCategory(tr("Global"), root_);
+  addShortcutType(EXIT_APPLICATION, global, tr("E&xit"), tr("Exit"));
+  addShortcutType(OPEN_FILE, global, tr("&Open..."), tr("Open file"));
+//  Uncomment once supported
+//  addShortcutType(NEW_FILE, global, tr("&New..."), tr("New File"));
+  addShortcutType(SHOW_DATABASE, global, tr("Show database view"));
+  addShortcutType(SHOW_LOG, global, tr("Show log view"));
+  addShortcutType(SHOW_OPTIONS, global, tr("&Options"), tr("Show the Dialog to select applications options"));
```
mkow: `dialog`

mkow: Also, these strings seem to occur twice in the code, please extract them to a constants and remove copy-paste ;)

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  addShortcutType(SHOW_SHORTCUT_OPTIONS, global, tr("Keyboard shortcuts"), tr("Show the Dialog to customize keyboard shortcuts"));
```
mkow: `dialog`

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  addShortcutType(VISUALIZATION_MANIPULATOR_FREE, three_d, tr("Free manipulator"), tr("Switch to free manipulator"));
+}
+
+ShortcutsItem* ShortcutsModel::addCategory(QString name, ShortcutsItem* parent){
+  auto item = new ShortcutsItem(name, parent);
+  parent->children.append(item);
+  return item;
+}
+
+ShortcutsItem* ShortcutsModel::addShortcutType(ShortcutType type, ShortcutsItem* parent, QString name, QString display_name) {
```
mkow: wrap

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+  ShortcutsItem* item;
+  if (display_name.isEmpty()) {
+    item = new ShortcutsItem(name, parent, type);
+  } else {
+    item = new ShortcutsItem(name, display_name, parent, type);
+  }
+  assert(!type_to_shortcut_.contains(type));
+  type_to_shortcut_[type] = item;
+  auto shortcuts = getShortcuts(type);
+  for (auto& shortcut : shortcuts) {
```
mkow: either `const auto&` or `auto`

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+    addShortcut(type, shortcut);
+  }
+  parent->children.append(item);
+  return item;
+}
+
+bool ShortcutsModel::checkIfConflicts(ShortcutsItem* item) const {
+  for (auto& shortcut : item->shortcuts()) {
```
mkow: ditto

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+
+void ShortcutsModel::updateShortcutsForType(ShortcutType type) {
+  auto item = type_to_shortcut_[type];
+  item->updateShortcutsForActions();
+  setShortcuts(type, item->shortcuts());
+}
+
+void ShortcutsModel::resetShortcutsForType(ShortcutType type) {
+  auto item = type_to_shortcut_[type];
+  for (auto shortcut : item->shortcuts()) {
```
mkow: const ref

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+    removeShortcut(type, shortcut);
+  }
+  for (auto shortcut : getShortcuts(type)) {
```
mkow: const ref

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+    addShortcut(type, shortcut);
+  }
+}
+
+void ShortcutsModel::resetShortcutsToDefault(ShortcutType type) {
+  auto shortcuts = defaultShortcuts();
+  auto item = type_to_shortcut_[type];
+  for (auto shortcut : item->shortcuts()) {
```
mkow: const ref

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+    removeShortcut(type, shortcut);
+  }
+  for (auto shortcut : shortcuts.value(type)) {
```
mkow: const ref

src/util/settings/shortcuts.cc
```diff
@@ -0,0 +1,473 @@
...
+    addShortcut(type, shortcut);
+  }
+}
+
+void ShortcutsModel::resetAllShortcutsToDefaults() {
+  for (auto iter : type_to_shortcut_) {
```
mkow: const ref ?

src/visualization/panel.cc
```diff
@@ -315,27 +319,33 @@ void VisualizationPanel::initOptionsPanel() {
   /////////////////////////////////////
   // Modes: digram / trigrams
   QColor icon_color = palette().color(QPalette::WindowText);
-  digram_action_ = new QAction(util::getColoredIcon(
-      ":/images/digram_icon.png", icon_color), tr("&Digram"), this);
+  digram_action_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::VISUALIZATION_DIGRAM, this,
+        util::getColoredIcon(":/images/digram_icon.png", icon_color), Qt::WidgetWithChildrenShortcut);
```
mkow: wrap

src/visualization/panel.cc
```diff
@@ -315,27 +319,33 @@ void VisualizationPanel::initOptionsPanel() {
...
   digram_action_->setToolTip("Digram Visualization");
   connect(digram_action_, SIGNAL(triggered()), this,
           SLOT(showDigramVisualization()));
 
-  trigram_action_ = new QAction(util::getColoredIcon(
-      ":/images/trigram_icon.png", icon_color), tr("&Trigram"), this);
+  trigram_action_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::VISUALIZATION_TRIGRAM, this,
+        util::getColoredIcon(":/images/trigram_icon.png", icon_color), Qt::WidgetWithChildrenShortcut);
```
mkow: wrap

src/visualization/panel.cc
```diff
@@ -362,10 +372,13 @@ void VisualizationPanel::initOptionsPanel() {
 
   /////////////////////////////////////
   // Sampling
-  QAction* show_more_options_action = new QAction(QIcon(":/images/more.png"),
-      tr("More options"), this);
+
+  QAction* show_more_options_action = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
```
mkow: wrap

src/visualization/trigram.cc
```diff
@@ -133,23 +133,20 @@ void TrigramWidget::prepareOptions(QMainWindow* visualization_window) {
   // Shape
   QToolBar* shape_toolbar = new QToolBar("Shape", this);
   shape_toolbar->setMovable(false);
-  cube_action_ = new QAction(this);
-  cube_action_->setIcon(
-      util::getColoredIcon(":/images/cube.png", icon_color, false));
+  cube_action_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::TRIGRAM_CUBE, this, util::getColoredIcon(":/images/cube.png", icon_color, false));
```
mkow: wrap

src/visualization/trigram.cc
```diff
@@ -133,23 +133,20 @@ void TrigramWidget::prepareOptions(QMainWindow* visualization_window) {
...
   connect(cube_action_, &QAction::triggered, std::bind(
       &TrigramWidget::setShape, this, EVisualizationShape::CUBE));
   shape_toolbar->addAction(cube_action_);
 
-  cylinder_action_ = new QAction(this);
-  cylinder_action_->setIcon(
-      util::getColoredIcon(":/images/cylinder.png", icon_color, false));
+  cylinder_action_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::TRIGRAM_CYLINDER, this, util::getColoredIcon(":/images/cylinder.png", icon_color, false));
```
mkow: wrap

src/visualization/trigram.cc
```diff
@@ -133,23 +133,20 @@ void TrigramWidget::prepareOptions(QMainWindow* visualization_window) {
...
   connect(cylinder_action_, &QAction::triggered, std::bind(
       &TrigramWidget::setShape, this, EVisualizationShape::CYLINDER));
   shape_toolbar->addAction(cylinder_action_);
 
-  sphere_action_ = new QAction(this);
-  sphere_action_->setIcon(
-      util::getColoredIcon(":/images/sphere.png", icon_color));
+  sphere_action_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::TRIGRAM_SPHERE, this, util::getColoredIcon(":/images/sphere.png", icon_color));
```
mkow: wrap

mkow:
> Ad 5 - which dialog are you talking about ? In the second dialog for choosing shortcut you can use Tab key to change focus, and then you can use buttons and/or Esc&enter/return (I thought it was fine enough). However in the first dialog I indeed by mistake removed ability to save changes using keyboard (you can still cancel using escape) - fixed.

That jumping with <kbd>Tab</kbd> to "Ok" button isn't too convenient and <kbd>Tab</kbd> and then <kbd>Enter</kbd>/<kbd>Esc</kbd> isn't too intuitive IMO. <kbd>Ctrl</kbd>+<kbd>Enter</kbd> is IMO a standard way for accepting dialogs/forms that capture plain <kbd>Enter</kbd>.

include/util/settings/shortcuts.h
```diff
@@ -0,0 +1,162 @@
...
+  virtual int columnCount(const QModelIndex& parent = QModelIndex()) const override;
+  virtual QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+  QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const override;
+  static ShortcutsModel* getShortcutsModel();
+  QAction* createQAction(util::settings::shortcuts::ShortcutType type, QObject* parent, Qt::ShortcutContext context = Qt::WindowShortcut);
+  QAction* createQAction(util::settings::shortcuts::ShortcutType type, QObject* parent, QIcon icon, Qt::ShortcutContext context = Qt::WindowShortcut);
+  void addShortcut(util::settings::shortcuts::ShortcutType type, QKeySequence& shortcut);
+  void removeShortcut(util::settings::shortcuts::ShortcutType type, const QKeySequence& shortcut);
+  ShortcutsItem* itemFromIndex(const QModelIndex& index) const;
+  QModelIndex indexFromItem(ShortcutsItem* item) const;
```
altran01: Due to how Qt interfaces are defined I would need to use const_cast somewhere if I add const here - I'm leaning towards leaving it as it is - how do you feel about it?

src/ui/shortcutedit.cc
```diff
@@ -0,0 +1,54 @@
...
+Qt::Key ShortcutEdit::key() {
+  return key_;
+}
+
+void ShortcutEdit::reset() {
+  key_ = Qt::Key_unknown;
+  clear();
+}
+
+void ShortcutEdit::keyPressEvent(QKeyEvent* event) {
```
altran01: This is part of Qt interface (in .h file you can see override).

src/ui/shortcutedit.cc
```diff
@@ -0,0 +1,54 @@
...
+  if (key == 0 || key == Qt::Key_unknown) {
+    // Unkown unhandled key pressed
+    return;
+  }
+  if (event->modifiers() != Qt::NoModifier && event->modifiers() != Qt::KeypadModifier) {
+    // Modfier pressed - we can't handle it correctly so we ignore it here
+    // and handle it through checkboxes
+    return;
+  }
+  if (key == Qt::Key_Shift || key == Qt::Key_Control || key == Qt::Key_Meta || key == Qt::Key_Alt) {
```
altran01: I wasn't able to break current checks on neither platforms I tested it on (including macOS) - AltGr always comes with modifiers, but I can include it if you will feel safer with it :)

src/ui/shortcutssettings.cc
```diff
@@ -0,0 +1,230 @@
...
+  ui->ctrlBox->setChecked(false);
+  ui->shiftBox->setChecked(false);
+  ui->metaBox->setChecked(false);
+  ui->lineEdit->reset();
+  ui->conflictLabel->clear();
+  disconnect();
+}
+
+QKeySequence ShortcutEditDialog::getSequence() {
+  int key = ui->lineEdit->key();
```
altran01: I would agree if not for the fact that Qt itself uses ints for it - so if it will bite us we won't be able to do anything about it either way (`key()` returns int) :P

src/ui/veles_mainwindow.cc
```diff
@@ -154,56 +155,70 @@ void VelesMainWindow::init() {
...
+        util::settings::shortcuts::SHOW_OPTIONS, this, Qt::ApplicationShortcut);
   options_act_->setStatusTip(
       tr("Show the Dialog to select applications options"));
   connect(options_act_, &QAction::triggered, this,
           [this]() { options_dialog_->show(); });
+
+  shortcuts_act_ = util::settings::shortcuts::ShortcutsModel::getShortcutsModel()->createQAction(
+        util::settings::shortcuts::SHOW_SHORTCUT_OPTIONS, this, Qt::ApplicationShortcut);
+  shortcuts_act_->setStatusTip(
+      tr("Show the Dialog to customize keyboard shortcuts"));
```
altran01: Its a copy paste from other dialog status tip :P

src/util/settings/shortcuts.cc
altran01: Right, my mistake.

src/util/settings/shortcuts.cc
altran01: In theory they could be different strings (though I used the same :P) - will do.

altran01: Hmm, where would you want me to put such strings @mkow?

altran01:
> That jumping with Tab to "Ok" button isn't too convenient and Tab and then Enter/Esc isn't too intuitive IMO. Ctrl+Enter is IMO a standard way for accepting dialogs/forms that capture plain Enter.

Done.

CMakeLists.txt
```diff
@@ -310,17 +310,17 @@ target_link_libraries(unpyc veles_db parser veles_network)
...
+  ${SRC_DIR}/ui/connectiondialog.ui
+  ${SRC_DIR}/ui/connectionnotificationwidget.ui
+  ${SRC_DIR}/ui/createchunkdialog.ui
+  ${SRC_DIR}/ui/databaseinfo.ui
+  ${SRC_DIR}/ui/gotoaddressdialog.ui
+  ${SRC_DIR}/ui/logwidget.ui
+  ${SRC_DIR}/ui/optionsdialog.ui
+  ${SRC_DIR}/ui/searchdialog.ui
+  ${SRC_DIR}/ui/shortcutssettings.ui
+  ${SRC_DIR}/ui/shortcutselection.ui
```
mkow: shortcutselection < shortcutssettings

include/util/settings/shortcuts.h
mkow: Ok, then leave it as is.

src/ui/shortcutedit.cc
mkow: ack

src/ui/shortcutedit.cc
mkow: Same for me, but let's stay on the safe side and check it, as the specification doesn't guarantee that it must come with other modifiers ;)

src/ui/shortcutssettings.cc
mkow: ack

src/util/settings/shortcuts.cc
mkow: kk, leave it as is

PR #219, opened 2017-05-23
Use argparse in Python server scripts

altran01:
Also allow to start server with in-memory database

### DIFF
--- a/python/plugin
+++ b/python/plugin
@@ -24,12 +24,13 @@ from veles.async_client.conn import AsyncRemoteConnection
 from veles.async_client.proto import create_unix_client, create_tcp_client
 from veles.util import helpers
 
-logging.basicConfig(level=logging.INFO)
-
-parser = helpers.get_client_argparse()
+parser = argparse.ArgumentParser(
+    parents=[helpers.get_client_argparse(), helpers.get_logging_argparse()])
 parser.add_argument('--plugin', action='append',
                     help='name plugin module to load')
 args = parser.parse_args()
+
+logging.basicConfig(level=logging.getLevelName(args.log_level))
 
 logging.info('Świtezianka plugin is starting up...')
 loop = asyncio.get_event_loop()
--- a/python/plugin
+++ b/python/plugin
@@ -14,31 +14,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import argparse
 import logging
 import asyncio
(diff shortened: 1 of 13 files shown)


python/srv.py
```diff
@@ -25,23 +25,34 @@
 
 logging.basicConfig(level=logging.INFO)
 
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    'database', help='path to database file, in-memory will be used if empty')
+parser.add_argument(
+    'url', help='either UNIX:<socket_path> or <ip>:<tcp port> to listen on')
```
wiktorn: Change part of the help to: `[<ip>:]<tcp port>` as this is current (desired?) behaviour

python/srv.py
```diff
@@ -25,23 +25,34 @@
 
 logging.basicConfig(level=logging.INFO)
 
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    'database', help='path to database file, in-memory will be used if empty')
```
wiktorn: Maybe we can move `database` to the end of the argument list and add `nargs='*'`so then it is no longer needed to supply empty string explicitly

python/srv.py
```diff
@@ -25,23 +25,34 @@
...
+parser.add_argument(
+    'url', help='either UNIX:<socket_path> or <ip>:<tcp port> to listen on')
+parser.add_argument(
+    'auth_key', help='hex-encoded up to 64bytes value that '
+                     'clients need to provide when connecting')
+parser.add_argument('plugin', nargs='*', help='name plugin module to load')
```
wiktorn: as `plugin` is optional, we can  change it to `--plugin` so it will not clash with database. Use `nargs=1, action='append'`

python/srv.py
altran01: Will do (but with `nargs='?'` :) ) - at first I didn't want to change it to avoid changing C++ code, but meh.

python/plugin
```diff
@@ -22,13 +22,13 @@ import signal
 
 from veles.async_client.conn import AsyncRemoteConnection
 from veles.async_client.proto import create_unix_client, create_tcp_client
+from veles.util import helpers
 
 logging.basicConfig(level=logging.INFO)
```
wiktorn: Please add argument to set logging level - also in proxy and srv.py

PR #220, opened 2017-05-25
Make hex-edit cursor movement configurable

altran01:


src/ui/hexedit.cc
```diff
@@ -222,6 +221,125 @@ HexEdit::HexEdit(FileBlobModel *dataModel, QItemSelectionModel *selectionModel,
   cursor_timer_.setInterval(700);
   cursor_timer_.start();
   connect(&cursor_timer_, &QTimer::timeout, this, &HexEdit::flipCursorVisibility);
+
+  createAction(util::settings::shortcuts::HEX_MOVE_TO_NEXT_CHAR, [this] () {
+    setSelection(current_position_ + 1, selection_size_);
+    if (!isByteVisible(current_position_)) {
+      scrollRows(1);
```
mkow: This seems to apply to almost all implementations of the actions in this file, please fix all of them.

src/ui/hexedit.cc
```diff
@@ -231,6 +349,13 @@ QModelIndex HexEdit::selectedChunk() {
   return chunkSelectionModel_->currentIndex();
 }
 
+void HexEdit::createAction(util::settings::shortcuts::ShortcutType type, const std::function<void ()>& f) {
+  auto action = ShortcutsModel::getShortcutsModel()->createQAction(
+        type, this, Qt::WidgetWithChildrenShortcut);
```
mkow: 6-space indent?

src/util/settings/shortcuts.cc
```diff
@@ -368,6 +390,30 @@ ShortcutsModel::ShortcutsModel() : root_(new ShortcutsItem()) {
...
+  addShortcutType(HEX_SELECT_NEXT_CHAR, hex_move, tr("Right with selecton"));
+  addShortcutType(HEX_SELECT_PREV_CHAR, hex_move, tr("Left with selection"));
+  addShortcutType(HEX_SELECT_NEXT_LINE, hex_move, tr("Down with selection"));
+  addShortcutType(HEX_SELECT_PREV_LINE, hex_move, tr("Up with selection"));
+  addShortcutType(HEX_SELECT_NEXT_PAGE, hex_move, tr("Page down with selection"));
+  addShortcutType(HEX_SELECT_PREV_PAGE, hex_move, tr("Page up with selection"));
+  addShortcutType(HEX_SELECT_START_OF_LINE, hex_move, tr("Start of line with selection"));
+  addShortcutType(HEX_SELECT_END_OF_LINE, hex_move, tr("End of line with selection"));
+  addShortcutType(HEX_SELECT_START_OF_DOCUMENT, hex_move, tr("Start of file with selection"));
+  addShortcutType(HEX_SELECT_END_OF_DOCUMENT, hex_move, tr("end of file with selection"));
```
mkow: `end` -> `End`

src/ui/hexedit.cc
mkow: Ok, I read the whole PR, seems that the bug was here before. CC: @adamiwaniuk

mkow:
<kbd>Esc</kbd> is wrongly marked as a conflict (collides with the same shortcut in 3D view), we should probably add some classes to the shortcuts and check conflicts only within them.

src/ui/hexedit.cc
adamiwaniuk: please create issue or something and after this change will be merged to master I will fix that

mkow: Done.
https://github.com/codilime/veles/issues/222

altran01:
> Esc is wrongly marked as a conflict (collides with the same shortcut in 3D view), we should probably add some classes to the shortcuts and check conflicts only within them.

Yeah it is wrongly marked - the problem is that it is not so easy to define such classes, and define with which other classes they conflict - it also depends a lot on how exactly our components bind actions to know whether they will conflict, or not (i.e. is the action bound to only one widget in dock, or to whole dock). Comparing with some IDE that I use:

- Qt Creator does the same thing as we do
- PyCharm doesn't mark conflicting shortcuts, but when adding new shortcut it does global conflict search
- Visual Studio - doesn't mark conflicting shortcuts, it has classes for shortcuts, but it still does global conflict search

src/ui/hexedit.cc
altran01: This is the default indent that Qt Creator creates in such situation (in fact in previous PR there was a lot of such indents) - what indent does Google C++ style guide use in such situation?

mkow: Seems I missed this in previous PRs ;) Should be 4 spaces, see https://google.github.io/styleguide/cppguide.html#Function_Calls

mkow:
Ok, let's leave it as is, it's not too important.

PR #223, opened 2017-05-29
Wip/altran/ssl

altran01:
WIP - at this moment it doesn't yet use the OpenSSL it compiles.
It drops support for MinGW, because it doesn't simply compile from Windows cmd, and instead requires use MSYS which seems to be counter productive - alternatively we can only remove MinGW from Jenkinsfile, while leaving it in other places with comment, that it doesn't work properly, but we wish luck in making it work.


CMakeLists.txt
```diff
@@ -418,6 +417,7 @@ add_executable(main_ui
 set_property(DIRECTORY APPEND PROPERTY ADDITIONAL_MAKE_CLEAN_FILES ${GENERATED_MOC_FILES})
 
 qt5_use_modules(main_ui Core Gui Widgets)
+add_dependencies(main_ui openssl)
```
adamiwaniuk: This dependency is just to test compilation, right ?

adamiwaniuk:
Copying OpenSSL dlls
Error: Target (for copy_if_different command) "Veles UI" is not a directory.
NMAKE : fatal error U1077: 'echo' : return code '0x1'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
10:07:46: The process "C:\Program Files\CMake\bin\cmake.exe" exited with code 2.
Error while building/deploying project VELES (kit: Desktop Qt 5.7.0 MSVC2015_64bit)
The kit Desktop Qt 5.7.0 MSVC2015_64bit has configuration issues which might be the root cause for this problem.
When executing step "Make"
10:07:46: Elapsed time: 08:21.

CMakeLists.txt
altran01: Yes I needed to add it somewhere to make sure it will compile. I'll move it to `veles_client` since it will be the place which uses ssl.

mkow:
Please mention perl dependency in README.rst

PR #227, opened 2017-06-02
focus fix

adamiwaniuk:


include/ui/dockwidget.h
```diff
@@ -90,6 +90,7 @@ class DockWidget : public QDockWidget {
   void splitVertically();
 
  protected:
+  void focusInEvent(QFocusEvent * event) Q_DECL_OVERRIDE;
```
altran01: nit. I think we decided to use `override` instead of `Q_DECL_OVERRIDE`.

adamiwaniuk: it is unifed with other methods in this file so I prefer to not change that, and some day change everything at once

PR #231, opened 2017-06-02
Move PROTO_VERSION from veles.server to veles.proto

koriakin:
Importing veles.server from veles.scli breaks Python 2 for obvious reasons.

Reported by mkow.


python/veles/proto/messages.py
```diff
@@ -22,6 +22,9 @@
 from veles.schema import model, fields
 from veles.schema.nodeid import NodeID
 
+# TODO generate this as a constant in C++
```
altran01: This TODO is fixed with this move.

koriakin: <3

PR #243, opened 2017-06-14
Fix Ctrl+C when running srv.py on Windows

mkow:


python/srv.py
```diff
@@ -77,5 +78,11 @@
     loop.add_signal_handler(signal.SIGINT, loop.stop)
 except NotImplementedError:
     pass
-loop.run_forever()
+if platform.system() == 'Windows':
```
altran01: Hmm, L77-80 is checking whether we have support for SIGINT handler (which decides if we need this hack or not). I think it would be better to merge it (and remember EAFP :) ).

PR #252, opened 2017-06-23
Fix hexedit cursor alignment (second try)

mkow:


src/ui/hexedit.cc
```diff
@@ -52,18 +52,20 @@ void HexEdit::recalculateValues() {
...
 
   // plus one for only addr
   rowsCount_ = (dataBytesCount_ + bytesPerRow_ - 1) / bytesPerRow_ + 1;
   rowsOnScreen_ =
       (viewport()->height() - horizontalAreaSpaceWidth_ * 2) / charHeight_;
 
-  spaceAfterByte_ = charWidht_ / 2;
-  hexAreaWidth_ = bytesPerRow_ * byteCharsCount_ * charWidht_ +
+  // Spacing needs to be odd for proper border alignment.
+  spaceAfterByte_ = (charWidth_ / 3 | 1);
```
adamiwaniuk: I will prefer to not use "|" here or at least please add parenthesis.

mkow: But what else could we use? Writing "/ 2 * 2 + 1" just to not use "|" doesn't make sense IMO.
Same with parentheses - '/' has very high priority; you don't add parentheses to "1 / 2 + 3" ;)

adamiwaniuk: I don't add parentheses to "1 / 2 + 3" because I know this priority from school, but people generally don't remember anything about priority of "|".

mkow: Hmm, I don't agree with this, because the rule is simple - all bit operators are below arithmetic operators. We should assume that the reader knows C++ enough and if not he can just google this.  Especially with division, which has a very high priority.

mkow: Still don't agree, but added them.

PR #280, opened 2017-07-14
Make HexEdit::paintEvent() 2x faster

mkow:
Changes:
* `painter.drawText()` -> `painter.drawStaticText()` + text cache (20% speedup)
* Removed unused `QSettings settings;` in a frequently called function (6% speedup)
* Changed one `i`-based loop to foreach (1.5%)
* Many other small changes, each giving few percents.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,9 @@ if(CMAKE_COMPILER_IS_GNUCXX)
   set(ADDITIONAL_LINK_LIBRARIES "pthread")
 endif()
 if(MSVC)
+  # Uncomment this line to use instrumentation in Visual Studio.
+  # set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /Profile")
+
   # set exception handling mode and linking mode
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc ")
   # build with multiple processes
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -6,3 +6,6 @@ CMakeLists.txt.user
 # Visual Studio files
 .vs
 CMakeSettings.json
+*.vsp
+*.vspx
+*.psess
```
wiktorn: Doesn't belong here

CMakeLists.txt
```diff
@@ -39,10 +39,13 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON)
 set(CMAKE_INSTALL_RPATH ".")
 if(CMAKE_COMPILER_IS_GNUCXX)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall --std=c++14") # All warnings
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14") # All warnings
```
wiktorn: this too

src/ui/hexedit.cc
```diff
@@ -114,8 +119,10 @@ HexEdit::HexEdit(FileBlobModel *dataModel, QItemSelectionModel *selectionModel,
       current_area_(WindowArea::HEX),
       cursor_pos_in_byte_(0),
       cursor_visible_(false),
+      selectionColor_(viewport()->palette().color(QPalette::Highlight)),
```
wiktorn: clang-tidy: `field 'cursor_visible_' will be initialized after field 'selectionColor_'` ?

catsuryuu: The issue is that order of initialization is different than order of fields in class.

mkow:
Fixed, please check again.

PR #287, opened 2017-07-24
Remove old db implementation.

koriakin:
A few pieces are still left -- these are used by current parsers.


mkow:
Please fix cmake, it doesn't build in current version.

PR #297, opened 2017-07-27
Use foreach in clang-format with long lists

wiktorn:
When argument list to clang-format is longer than 8192 (limit for cmd.exe) on
Win32 platform use multiple targets and format files one by one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -525,10 +525,10 @@ if(CLANG_TIDY)
 
   # foreach is needed so maximum command line length is not reached
   # create intermediary targets to allow parallel checks with -jN
+  set(LINT_INDEX 0)
   foreach(LINT_SOURCE_FILE ${LINT_ALL_SOURCE_FILES})
-    list(APPEND LINT_ADDED_TARGETS ${LINT_SOURCE_FILE})
-    list(LENGTH LINT_ADDED_TARGETS LINT_ADDED_TARGETS_LEN)
-    set(LINT_TARGET_NAME "lint-${LINT_ADDED_TARGETS_LEN}")
+    set(LINT_TARGET_NAME "lint-${LINT_INDEX}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -561,16 +561,39 @@ find_program(CLANG_FORMAT NAMES clang-format clang-format-4.0 clang-format-3.9 c
...
-      ${CLANG_FORMAT}
-      -style=file
-      -sort-includes
-      -i
-      ${LINT_ALL_SOURCE_FILES}
-  )
+  file(GLOB_RECURSE FORMAT_ALL_SOURCE_FILES src/*.cc include/*.h)
+  string(LENGTH "${FORMAT_ALL_SOURCE_FILES}" FORMAT_ALL_SOURCE_FILES_LEN)
+  if(WIN32 AND FORMAT_ALL_SOURCE_FILES_LEN GREATER 8192)
+    # use multiple targets to overcome cmd 8192 length limit
```
mkow: Seems it's 8191 in fact. See https://support.microsoft.com/en-us/help/830473/command-prompt-cmd--exe-command-line-string-limitation.

CMakeLists.txt
```diff
@@ -561,16 +561,39 @@ find_program(CLANG_FORMAT NAMES clang-format clang-format-4.0 clang-format-3.9 c
...
+    add_custom_target(format)
+
+    foreach(FORMAT_SOURCE_FILE ${FORMAT_ALL_SOURCE_FILES})
+      list(APPEND FORMAT_ADDED_TARGETS ${FORMAT_SOURCE_FILE})
+      list(LENGTH FORMAT_ADDED_TARGETS FORMAT_ADDED_TARGETS_LEN)
```
mkow: Why not an int variable? Seems that only length of FORMAT_ADDED_TARGETS is used.

PR #302, opened 2017-08-01
Read many bytes at once from edit engine

catsuryuu:


mkow:
`fix clang format`
Please squash this or rename, you aren't fixing clang-format, but code formatting with help from clang-format (current commit message sounds like you are fixing `format` target in our cmake)

PR #314, opened 2017-09-01
Wip/catsuryuu/modifications highlighting

catsuryuu:
Bytes marked as modified are exactly the same as these that will be send to server.

### DIFF
--- a/include/util/edit.h
+++ b/include/util/edit.h
@@ -48,9 +48,7 @@ class EditEngine {
   void clear();
 
   uint64_t byteValue(size_t pos) const;
-  uint64_t originalByteValue(size_t pos) const;
   data::BinData bytesValues(size_t pos, size_t size) const;
-  data::BinData originalBytesValues(size_t pos, size_t size) const;
   QVector<bool> modifiedPositions(size_t pos, size_t size) const;
 
  private:
--- a/include/util/edit.h
+++ b/include/util/edit.h
@@ -18,9 +18,9 @@
 
 #include <cstdlib>
 #include <memory>
+#include <vector>
 
 #include <QMap>
-#include <QVector>
 
 #include "data/bindata.h"
 #include "ui/fileblobmodel.h"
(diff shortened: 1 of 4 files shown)


src/util/edit.cc
```diff
@@ -176,8 +172,46 @@ data::BinData EditEngine::bytesValues(size_t pos, size_t size) const {
...
+          result[i] = true;
+        }
+      }
+      ++it;
+      ++next_it;
+    }
+    if (it->fragment_ != nullptr) {
+      for (size_t i = it.key() - pos; i < size; ++i) {
+        result[i] = true;
+      }
```
mkow: Please deduplicate code here. For example, you can fold the for-loops into a lambda taking (start, end) as arguments.

catsuryuu: DONE

catsuryuu: Fixed

include/util/edit.h
```diff
@@ -64,6 +63,9 @@ class EditEngine {
   };
 
   ui::FileBlobModel* original_data_;
+  // TODO(catsuryuu): change to std::map after switching to C++17 (and use
+  // `insert_or_assign`)
+  // Reason: QMap works on `int` type
```
mkow: 'QMap works on `int` type' -> 'QMap uses `int` for container size'
Current version may be unclear for a reader, especially that in the next line you use `size_t`.

src/util/edit.cc
```diff
@@ -175,8 +171,38 @@ data::BinData EditEngine::bytesValues(size_t pos, size_t size) const {
...
+std::vector<bool> EditEngine::modifiedPositions(size_t pos, size_t size) const {
+  std::vector<bool> result(size);
+
+  size_t end_pos = pos + size;
+  auto next_it = address_mapping_.upperBound(pos);
+  assert(next_it != address_mapping_.cbegin());
+  auto it = next_it;
+  --it;
+
+  auto setResultTrue = [&it, &result](size_t start, size_t end) {
```
mkow: it's a local variable, so it should be named `set_result_true`

include/util/edit.h
catsuryuu: Fixed

PR #329, opened 2017-09-14
Add remapOrigin in EditEngine

catsuryuu:
`remapOrigin` updates the address mapping after remapping request from server.

### DIFF
--- a/src/util/edit.cc
+++ b/src/util/edit.cc
@@ -38,7 +38,7 @@ void EditEngine::changeBytes(size_t pos, const data::BinData& bytes,
 
   has_changes_ = true;
 
-  size_t end_pos = pos + bytes.size();
+  const size_t end_pos = pos + bytes.size();
   auto next_it = address_mapping_.upperBound(pos);
   assert(next_it != address_mapping_.cbegin());
   auto it = next_it;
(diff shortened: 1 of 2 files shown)


src/util/edit.cc
```diff
@@ -116,6 +116,59 @@ void EditEngine::removeBytes(size_t pos, size_t size, bool add_to_history) {
...
+      }
+      local_pos = it.key() + (origin_pos - it->offset_);
+    }
+    ++it;
+  }
+
+  const size_t origin_end_pos = origin_pos + old_size;
+  size_t local_end_pos = origin_end_pos;
+
+  while (it != address_mapping_.cend()) {
```
mkow: duplicated code, please use a lambda

src/util/edit.cc
```diff
@@ -253,19 +302,22 @@ data::BinData EditEngine::getDataFromEditNode(const EditNode& edit_node,
 void EditEngine::remap(size_t pos, size_t old_size, size_t new_size) {
   // This variable can be overflowed when `old_size` > `new_size`,
   // but it will be OK when we add this difference to some address.
-  size_t offset = new_size - old_size;
+  const size_t offset = new_size - old_size;
+  // remap is O(n), so if `offset == 0` then we should use something faster
```
mkow: It's not clear what `n` means here. Change size in bytes? Size of the mapping? (*I* know what it means, but other readers probably don't).

src/util/edit.cc
```diff
@@ -253,19 +302,22 @@ data::BinData EditEngine::getDataFromEditNode(const EditNode& edit_node,
...
+  // (i.e. `changeBytes`)
```
mkow: you mean `e.g.`?

mkow: Why do we crash in this case instead of using slower algorithm?

PR #353, opened 2017-09-27
Calculate nice gl_PointSize in trigram view.

### LINKED ISSUE #343: Point scaling dependent on viewport size in 3D.

koriakin:
Fixes #343.

### DIFF
--- a/src/visualization/shaders/trigram/fshader.glsl
+++ b/src/visualization/shaders/trigram/fshader.glsl
@@ -17,8 +17,13 @@
 #version 330
 
 in float v_pos;
+in float v_factor;
 layout (location = 0, index = 0) out vec4 o_color;
 uniform float c_brightness;
 void main() {
-        o_color = vec4(1.0 - v_pos, 0.5, v_pos, 1) * c_brightness;
+	vec2 diff = vec2(dFdx(gl_PointCoord.x), dFdy(gl_PointCoord.y));
+	float cdist = length(gl_PointCoord.xy - vec2(0.5, 0.5)) - length(diff) / 2;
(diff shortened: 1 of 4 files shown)


src/visualization/shaders/trigram/vshader.glsl
```diff
@@ -19,7 +19,9 @@
 uniform mat4 xfrm;
 uniform usamplerBuffer tx;
 uniform uint sz;
-uniform float c_cyl, c_sph, c_pos;
+uniform float tz;
```
mkow: what is `tz`? could you find a name which is easier to decode? (or comment it)

src/visualization/shaders/trigram/fshader.glsl
```diff
@@ -20,5 +20,7 @@ in float v_pos;
 layout (location = 0, index = 0) out vec4 o_color;
 uniform float c_brightness;
 void main() {
-        o_color = vec4(1.0 - v_pos, 0.5, v_pos, 1) * c_brightness;
+        float cdist = length((gl_PointCoord.xy - vec2(0.5, 0.5)));
```
mkow: Redundant parentheses?

src/visualization/shaders/trigram/vshader.glsl
```diff
@@ -19,7 +19,9 @@
...
+uniform float c_cyl, c_sph, c_pos, c_ort;
+uniform float ptsz_factor;
```
mkow: `point_size_factor` would be better IMO, it's easier to read and it's used only in two places, so the lengthiness is not annoying too much.

PR #372, opened 2017-11-03
Add Qt Creator files to gitignore

prybicki:


mkow:
How did you produce any .veles file using Qt Creator? AFAIR it didn't generate anything like this, at least for me.
Also, adding such a broad wildcard to the gitignore sounds like a bad idea to me.

PR #376, opened 2017-11-14
New readme

### LINKED ISSUE #360: Lacking Ubuntu dependencies
> python3-msgpack
> python3-openssl
I installed all dependencies from README for Ubuntu. Without foregoing two packets Veles doesn't work (only without venv).

mkow:
I refreshed our readme a bit by splitting it into files and adding more detailed build instructions. I also converted everything into MD format, because it seems that GitHub's implementation of RST can't highlight code.

I'm also trying to find a good screen for the main readme, currently I have two candidates:
- https://github.com/codilime/veles/tree/9bb8edbd6624b7cde07194a2d9d9aba015acdc40
- https://github.com/codilime/veles/tree/e7324b1d6a8f6ba9c23f40cf41e7b2d8e53c7e46

If you have an idea for a better screenshot please go ahead and paste it in the comments :)

Fixes #360.


BUILDING.md
```diff
@@ -0,0 +1,189 @@
...
+## Linux
+
+Dependencies:
+
+- `cmake` >= 3.1.0
+- `zlib`
+- `qt` >= 5.5 ( >= 5.6 is required if you want to rearrange tabs using
+drag&drop)
+- `Python3.5+`
+    - `virtualenv`
```
koriakin: Should not be necessary (python 3.5 has a venv module bundled, which supersedes virtualenv)

BUILDING.md
```diff
@@ -0,0 +1,189 @@
...
+- `python3-msgpack`
+- `python3-openssl`
```
koriakin: IIRC it's called python3-pyopenssl

BUILDING.md
```diff
@@ -0,0 +1,189 @@
...
+- `OpenSSL` >= 1.0.0, < 1.1.0
+- `clang-tidy`
```
koriakin: These 3 are optional

fenollp:
I like the first one :) Do you mind sharing the BMP with me?
As a 3rd candidate [you may like this one](https://github.com/fenollp/voidstar/blob/master/data/BigPictureBG.tga)

mkow:
Thanks for the feedback :)
I converted [this](http://www.tapeciarnia.pl/tapety/normalne/213236_laka_maki_drzewo_tecza_wiosna.jpg) to a 24-bit BMP and opened the result in Veles.

Regarding your image: looks much better than the first one on the trigram view, but it's not "interesting" on the minimap (the same visual pattern repeats for the whole file). Maybe I'll try to find some ROM dump with images embedded inside, so there would be different data types packed inside (you can also try if you have any dumps). Otherwise I could just make a .tar file with your image and some other stuff, so it would look better.

fenollp:
Thanks :) I found some weird squid shape near the end of this image after converting it to TGA with gimp!
Anyway yeah game assets are a great resource, looking forward to see what you'll pick.

mkow:
@fenollp I packed few random files into a tarball, results:
https://github.com/codilime/veles/tree/0fffd2b0eebd3ec03ebc367c8c381b5a3dbf7f3e

BUILDING.md
```diff
@@ -0,0 +1,191 @@
...
+- `python3-msgpack`
+- `python3-openssl` or `python3-pyopenssl` (depending on your distro)
+- `OpenSSL` >= 1.0.0, < 1.1.0
+
+Optional dependencies: (recommended for developing)
+- `clang-tidy`
+- `clang-format`
+- `libclang-common-dev`
+
+Optional dependencies needed for running tests:
```
koriakin: Squash with previous list

BUILDING.md
```diff
@@ -0,0 +1,191 @@
...
+If you want to install to a non-default directory, you'll also need to pass
+it as an option to cmake before building, eg.:
+```bash
+$ cmake -D CMAKE_INSTALL_PREFIX:PATH=/usr/local ..
+```
+
+Python tests dependencies:
+
+- `git`
+- `Python2.7`
```
koriakin: unfortunately, tox currently needs exactly py3.5 to work (3.6+ won't do) -- needs to be mentioned here or fixed

BUILDING.md
```diff
@@ -0,0 +1,191 @@
...
+
+You will need to install the following:
+
+- `Qt` for `msvc2017_64`
+- `Visual Studio 2017` (Community edition is enough; you will need MSVC even
+when using only Qt Creator)
+- `cmake` + `git` (both can be installed with Visual Studio)
+- `perl` (only for building OpenSSL; e.g.
+[Strawberry Perl](http://strawberryperl.com/))
+- `Python3.5+` (only for running Veles server; you will also need to install
```
koriakin: "Only for running server" makes it sound like server is optional

PR #379, opened 2017-11-29
Remove empty space below data in hexedit. Change scrolling.

### LINKED ISSUE #375: Hexview should render as many rows as possible
Currently, there's a horizontal line which could be replaced by displaying additional bytes.
Example

chivay:
Fixes #375

### DIFF
--- a/include/ui/hexedit.h
+++ b/include/ui/hexedit.h
@@ -102,9 +102,13 @@ class HexEdit : public QAbstractScrollArea {
   qint64 startOffset_;
   /** Total number of rows in hex edit (counting last address only row) */
   qint64 rowsCount_;
+  /** Total number of rows in hex edit (not counting address only row) */
+  qint64 data_rows_count_;
   /** Number of rows displayed on the screen (calculated from window and font
    * height) */
   qint64 rowsOnScreen_;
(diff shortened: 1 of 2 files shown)


mkow:
I'm also not sure whether we want that longer scroll thing. I'm used to jumping to the end of file using the scrollbar, which in this version results in seeing only the last row. Any good use-case in favor of keeping this change?

include/ui/hexedit.h
```diff
@@ -105,6 +105,7 @@ class HexEdit : public QAbstractScrollArea {
   /** Number of rows displayed on the screen (calculated from window and font
    * height) */
   qint64 rowsOnScreen_;
+  qint64 totallyVisibleRows_;
```
mkow: Should be `totally_visible_rows_`, unfortunately the old code doesn't use Google style (I'll refactor it someday).

mkow:
Interesting: it works well now when I press Ctrl+End, but not when I press Ctrl+Shift+End (the last line is cropped by few pixels in the latter case).

mkow:
Same for manually moving cursor downwards.

chivay:
Everything seems to be working right now.

PR #382, opened 2017-12-02
FreeBSD support

lepot311:
Tested on FreeBSD 11.0-RELEASE


CMakeLists.txt
```diff
@@ -82,6 +82,14 @@ if(WIN32)
   set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL ${WINDOWS_APPLICATION_OUT_DIRECTORY})
 endif()
 
+# FreeBSD support
+
+# link BSD threading
+if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
```
mkow: I think it should be `STREQUAL` instead of `MATCHES` (the latter is for regexes).

src/ui/connectionmanager.cc
```diff
@@ -220,6 +220,9 @@ void ConnectionManager::startLocalServer() {
   if (!check_file.exists()) {
     python_interpreter_executable = QString("python3");
   }
+#elif defined(Q_OS_FREEBSD)
+  QString python_interpreter_executable(
+   "/usr/local/bin/python3");
 #else
 #warning \
     "This OS is not officially supported, you may need to set this string manually."
```
mkow: Hmm, the `#else` branch here seems to lack variable definition (`QString` before `python_interpreter_executable`). If you want to can also fix this in this PR ;)

CMakeLists.txt
```diff
@@ -82,6 +82,14 @@ if(WIN32)
   set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL ${WINDOWS_APPLICATION_OUT_DIRECTORY})
 endif()
 
+# FreeBSD support
+
+# link BSD threading
```
mkow: Could you please move the comment inside the `#if`? Also, starting the comment with a capital letter IMO looks better (like in other places in this file) ;)

mkow:
Unfortunately, I couldn't make GitHub link my fixup commit into this PR, so I opened a new one (#388) and will close this.

PR #383, opened 2017-12-03
Make trigram colors changeable in trigram class.

koriakin:


include/ui/color_picker_button.h
```diff
@@ -0,0 +1,43 @@
...
+
+class ColorPickerButton : public QPushButton
+{
+  Q_OBJECT
+
+ public:
+  ColorPickerButton(const QColor& initial_color, QWidget* parent = nullptr);
+  QColor getColor() const;
+
+ private slots:
```
koriakin: Future-proofing: shouldn't this be a public function, to allow changing the color from outside?

mkow: Sure, fixed.

PR #386, opened 2017-12-06
Fix highlighting item on hover in database tab.

### LINKED ISSUE #330: Element under mouse is not highlighted in dark theme in Database tab
To see the difference you can compare behavior of Database tab in white and dark themes.

mwarzynski:
Fix #330


mwarzynski:
Done.

PR #387, opened 2017-12-06
Animate reference frame axes smoothly on coordinate system change.

koriakin:


src/visualization/shaders/trigram/rf_vshader.glsl
```diff
@@ -3,7 +3,17 @@ in vec3 vert;
 in vec3 color;
 out vec3 int_color;
 uniform mat4 mvp;
+uniform float c_cyl, c_sph;
+const float TAU = 3.1415926535897932384626433832795 * 2;
 void main() {
-  gl_Position = mvp * vec4(vert, 1.0);
+  vec3 xpos = vert * vec3(2, 2, 2) - vec3(1, 1, 1);
```
mkow: Could you add some comments to this function to explain what happens here?

Issue #390, opened 2017-12-13
Add a unified way to reload settings for all our UI classes


mkow:
Two implementation ideas:
- virtual method (reapplySettings) in a common class
- global object which generates a reapplySettings signal, so other windows can subscribe it

Issue #389, opened 2017-12-10
The options dialog needs an "apply" button


koriakin:
Too many steps needed right now to check out 3D color schemes.

PR #395, opened 2018-01-12
Add "Apply" button to the Options dialog

mkow:
I also refactored a bit the way we reload settings.

### DIFF
--- a/include/ui/velesapplication.h
+++ b/include/ui/velesapplication.h
@@ -1,0 +1,42 @@
+/*
+* Copyright 2018 CodiLime
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+*/
+#pragma once
+
+#include <QApplication>
+
+namespace veles {
+namespace ui {
+
+// A modified QApplication which stores an additional global state.
+class VelesApplication : public QApplication {
+  Q_OBJECT
+
+ public:
+  using QApplication::QApplication;
+
+  static VelesApplication* instance() {
+    // This is always a proper cast, because there's only one instance of
+    // `QCoreApplication` (created in `main`) which is `VelesApplication`.
+    return static_cast<VelesApplication*>(QCoreApplication::instance());
+  }
+
+ signals:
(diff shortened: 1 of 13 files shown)


mkow:
Fixes #389.

koriakin:
Are you sure using an interface is the right idea? It seems to be a perfect opportunity to use signals:

- have a top-level settings object with a "settings changed" signal
- have the options dialog emit that signal
- in the constructor of every settings-using widget, connect the signal to an internal slot

mkow:
I'm afraid that this solution only sounds good, but in practice will be much less nice.
Reasons:
* Non-POD globals in C++ are dangerous in general (e.g. GCS forbids them: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables)
* Global QObjects are even more dangerous (e.g. because they outlive the main QApplication instance). Some things can probably break during application exit.

I also don't like the idea of scanning all widgets and calling a method on them, but this still seems to be the safest solution to me. So, in general, on the high-level I prefer your solution because it seems cleaner, but it seems that C++ will make it hard to implement safely.

Other solution would be to define a signal in `MainWindowWithDetachableDockWidgets` and emit it for every main window, but then we assume that all of our main windows derive from `MainWindowWithDetachableDockWidgets`, which probably won't hold in the future, and causes a lot of troubles when moving widgets between main windows.

malpunek:
There might be a solution for static non-PODs here - https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use

Apart from that I don't really see why `options` should be static. Why couldn't we subclass `QApplication` to include extra members like `private Options options; public signal optionsChanged();`? Or maybe we can treet `options` as a special kind of applications style (QStyle) and then simply use `QApplication::setStyle(options)`?

include/ui/velesapplication.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace veles {
+namespace ui {
+
+// A modified QApplication which stores an additional global state.
+class VelesApplication : public QApplication {
+  Q_OBJECT
+
+public:
+  using QApplication::QApplication;
```
malpunek: Is this necessary?

src/ui/optionsdialog.cc
```diff
@@ -79,20 +82,17 @@ void OptionsDialog::accept() {
       color_3d_begin_button_->getColor());
```
malpunek: I know this is not your code, but shouldn't these be slots of HexEdit & Visualization connected to settingsChanged signal?

include/ui/velesapplication.h
mkow: Yes. This line inherits a constructor which is used in main().

src/ui/optionsdialog.cc
mkow: No, as explained in private.

malpunek:
Czy to nie fixuje czasem #390 ?

mkow:
Right :)

Fixes #390.

PR #396, opened 2018-01-15
Code improvements in EditEngine

catsuryuu:
- Remove unnecessary `find`.
- Now all entirely overlapped nodes are removed in `while` loop (the last one too). `last_modified_node` will be never created unnecessarily.

### DIFF
--- a/src/util/edit.cc
+++ b/src/util/edit.cc
@@ -70,8 +70,9 @@ void EditEngine::modifyBytes(size_t pos, const data::BinData& bytes,
   if (next_it == address_mapping_.cend() || end_pos <= next_it.key()) {
     // This is the case when whole query range is located in only one node.
     if (it->fragment_ == nullptr) {
-      // TODO(catsuryuu): little refactoring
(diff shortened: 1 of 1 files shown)


src/util/edit.cc
```diff
@@ -70,8 +70,9 @@ void EditEngine::modifyBytes(size_t pos, const data::BinData& bytes,
   if (next_it == address_mapping_.cend() || end_pos <= next_it.key()) {
     // This is the case when whole query range is located in only one node.
     if (it->fragment_ == nullptr) {
-      // TODO(catsuryuu): little refactoring
-      if (address_mapping_.find(end_pos) == address_mapping_.cend()) {
+      if (next_it == address_mapping_.cend() || end_pos != next_it.key()) {
```
catsuryuu: Should I use `<` instead of `!=` for readability?

mkow: Yup, `<` seems to be slightly better, but I don't have any strong preference here.

PR #404, opened 2018-01-26
Fix cursor position in server url

### LINKED ISSUE #377: Fix "Server url" field in "Connection" dialog
After typing anything not at the end of a text, "Server url" field in Connection Dialog moves cursor to the end pos. It's very hard to type something in the middle of already written text.

boryspoplawski:
Fixes #377


src/ui/dialogs/connectiondialog.cc
```diff
@@ -100,7 +100,16 @@ ConnectionDialog::ConnectionDialog(QWidget* parent)
   connect(ui_->server_url_line_edit, &QLineEdit::textEdited,
           [this](const QString& text) {
             QString trimmed_text = text.trimmed();
+            int pos = ui_->server_url_line_edit->cursorPosition();
+
+            if (0 < trimmed_text.length()) {
```
mkow: https://en.wikipedia.org/wiki/Yoda_conditions

PR #447, opened 2019-06-16
Lift the default size limit of 1MB that msgpack imposes on binary data.

yurivict:


parkerlreed:
Can confirm this works here (opening a 13MB binary)

